PR #10, opened 2023-02-19
5 implement imm

GidonFrischkorn:
I have added the model implementation and the formula parsing for the IMM to the `fit_model` function. I implemented all three versions of the IMM proposed by Oberauer et al. (2017). That is the full IMM, the IMMabc without assuming swaps depending on spatial distance to the target, and the IMMbsc assuming that swaps can only occur as a function of spatial proximity to the target.

Would be great if you have a look over the changes and merge this branch to the master, once you checked everything.


venpopov:
Looks good, but before I merge, can you fix these two things?

line 34: You refer to parameter `spaDist` but later in the code it is `spaPos` everywhere. So fix it to spaPos as well? 
line 456: add the new models in the error message for incorrect argument

GidonFrischkorn:
I included the changes. Hope everything is alright now.

venpopov (approved):

Issue #39, opened 2024-02-01
Add a utility helper function add_model()


venpopov:
@GidonFrischkorn 's idea - to have and add_model() function that appends all the components necessary for creating a new model to the relevant files...

- e.g. check_data.newmodel, configure_model.newmodel(), .model_newmodel() class, etc
- suggest a vignette with explanation

PR #47, opened 2024-02-02
Move model specific functions to individual files

### LINKED ISSUE #46: move model-specific functions to individual files
move model-specific functions to individual files so that adding new models would be isolated from the current code

venpopov:
#### Summary

Models and their S3 methods now live in individual files of the form `bmm_model_modelname.R`. This way adding new models will not touch existing files.

Closes #46 and #39 

Added a use_model_template() function, which will create a new file with a skeletton for all the components necessary for new models.

#### Tests

#### Release notes

- Developers can now use the use_model_template() function to create a template file with all components necessary for defining new models


GidonFrischkorn (approved):
These changes all look good to me. Thanks for tidying this up.

PR #54, opened 2024-02-06
Add get_model_prior function

GidonFrischkorn:
This commit adds the get_model_prior function. Currently there is no post processing of the priors implemented. So, all priors will be provided even if the parameters will be internally fixed.

#### Tests


R/fit_model.R
```diff
@@ -9,13 +9,8 @@
...
 #' @param data An object of class data.frame, containing data of all variables
-#'   used in the model. Response, target and lure values must be in radians.
-#'   There must be as many lure value columns as the maximum setsize-1. For
-#'   setsizes smaller than the maximum, values for non-existing non_targets must
-#'   be coded as NA. The outcome variable must be response error relative to the
-#'   target, not the raw response. Similarly, the lure values must be coded
-#'   relative to the target. If the lure values are absolute, you must subtract
-#'   from them the value of the target before running the model
+#'   used in the model. The names of the variables must match the variable names
+#'   passed to the `bmmmodel` object for required argurments.
```
venpopov, line 13: good catch

venpopov:
Looks good and it runs for me. Just one question - it currently returns the default brms priors, but it does not take into account the priors that we set ourselves. This is because brms::get_prior only returns the brms priors. Is this intended? Ideally we would like to return the priors set by brms plus ours

GidonFrischkorn:
Yes, that is something that we would need to do in the post processing. Either we can just return the priors for parameters that are linked to parameters in the model that are predicted and not fixed to specific values. Or we could return the default priors that we specify for each model.

Do you have a preference for one of the options? My first hunch is to just return priors for the parameters of the models as these are the only ones that users should think about specifying themselves, right?

venpopov:
It's actuall a simple change to combine the brms default priors with ours - with a small change to the combine_prior function. I actually just implemented it and will push to the same branch and you can review

GidonFrischkorn:
Perfect. I will have a quick look at it and if there are no further issues you can approve the pull request and we merge this into develop.

venpopov:
ok, check it out

GidonFrischkorn:
Nice, thanks for the additional commits and adding the unit tests. From my side this looks good :-)

GidonFrischkorn:
I think you still have to approve the pull request so that we can merge it.

venpopov (approved):

PR #74, opened 2024-02-08
Update data check and add deg2rad & rad2deg

GidonFrischkorn:
#### Tests

#### Release notes
- Add deg2rad & rad2deg functions
- Only throw warning for VWM models if values larger than 2pi are detected for the response variable


venpopov (approved):
I added this line to the warning:

The model requires the response variable to be in radians.\n

so that users don't wonder what to do

PR #76, opened 2024-02-09
Update README

GidonFrischkorn:
Update the README of the bmm package to reflect the new structure of the package and refer to the vignettes for in-depth explanations for fitting the different models.

#### Tests


GidonFrischkorn:
I have not yet found a way to automatically link the vignettes. Given that you might have had a look into this for the pkgdown website. I thought it might be easier for you to implement that.

Issue #79, opened 2024-02-09
require dependent variable to be passed to the model to avoid specifying it in the formula


GidonFrischkorn:
- dv as required argument
- add option to predict the location of the target distribution in the mixture models

venpopov:
one solution to the problem you told me at lunch would be write a bmm formula function similar to brms::bf, in which we have more flexibility, though I suspect this might be a lot of work and not sure if it's worth it. Just an idea

venpopov:
although maybe we should stick with the current syntax to be consistent with brms

GidonFrischkorn:
Yes, I thought a bit about possible solutions as well, but did not come up with one yet. I will have to look into this in more detail for the ezDM implementations anyway. But for now, I would keep the formula syntax the way it is.

GidonFrischkorn:
I have looked a bit into the object created by the `brmsformula` function. And think there might be an easier solution than initially thought to this problem. I will explore it a little, but in short. The `bf` function separates the first formula as the "main" formula and stores the rest of the formulas in an additional list labeled `pforms`. I assume this is an abbrehviation for parameter formulas.

I think it should be possible to write a simple `bmm_formula` function that creates this list taking in the same type of formulas that the `brmsformula` function takes in. And then we can actually just add the correctly formatted list into the `formula$pforms` list. 

I will check this out at the beginning of next week and keep you updated.

venpopov:
that's along the lines of what I was thinking above. Glad to hear it is simpler than expected!

GidonFrischkorn:
I had a bit of time to look into this and thing I should be able to provide a solution. This also entails checking if formulas for all model parameters were specified and throwing a warning if this is not the case while only estimating a fixed intercept for them.

I will keep you posted on the progress and then we can see, if the solution seems feasible for all models and also in the long run.

But for some models, such as The ezDM, I will certainly use this functionality. So, I will keep exploring this anyway.

PR #92, opened 2024-02-12
add DV as variable to model() call & add bmm_formula

### LINKED ISSUE #98: Proposal to rename imm/3p arguments
I've been thinking about this for a while, and now is a good time to bring it up - to me spaPos is unintuitive as an argument. Without reading the documentation it sounds like it asks for the position of non_targets, whereas it asks for the distance. 

Can I suggest renaming this to spa_dist? More generally, both non_targets and spaPos refer to different features of the non_targets relative to the target. So mayb we should consider something like nt_features/nt_values and nt_dists. That is, I'm suggesting renaming:

`non_targets` to `nt_features` or `nt_values`
`spaPos` to `nt_dists`

makes it consistent with the parameter naming (thetant), and makes it clear what type of information each contains and how they are related.  Alows opens up the positibility to expand it naturally in the future if we want to allow people to specify multiple distance metrics via nt_dists1, nt_dists2, ntdists3, etc... The model doesn't care what they represent and it would be up to the user to interpret within the context of whatever distance they are using. What do you think?

GidonFrischkorn:
#### Summary

This commit moves away from using brmsformulas to be supplied to the fit_model function. Additionally, the dependent varible needs to be specified in the model call and is no longer part of the model formula. 

Closes #98. 

#### Tests

### New features

* BREAKING CHANGE: The `fit_model` function now requires a `bmmformula` to be passed. The  syntax of the `bmmformula` or its short form `bmf` is equal to specifying a `brmsformula`. However, as of this version the `bmmformula` only specifies how parameters of a `bmmmodel` change across experimental conditions or continuous predictors. The response variables that the model is fit to now have to be specified when the model is defined using `model = bmmmodel()`. (#79)
* BREAKING CHANGE: The `non_target` and `spaPos` variables for the `mixture3p` and `IMM` models were relabled to `nt_features` and `nt_distances` for consistency. This is also to communicate that distance is not limited to spatial distance but distances on any feature dimensions of the retrieval cues. Currently, still only a single generalization gradient for the cue features is possible. 
* This release includes reference fits for all implemented models to ensure that future changes to the package do not compromise the included models and change the results that their implementations produce.
* The `check_formula` methods have been adapted to match the new `bmmformula` syntax. It now evaluates if formulas have been specified using the `bmmformula` function, if formulas for all parameters of a `bmmmodel` have been specified and warns the user that only a fixed intercept will be estimated if no formula for one of the parameters was provided. Additionally, `check_formula` throws an error should formulas be provided that do not match a parameter of the called `bmmmodel` unless they are part of a non-linear transformation.
* You can now specify formulas for internally fixed parameters such as `mu` in all visual working memory models. This allows you to predict if there is response bias in the data. If a formula is not provided for `mu`, the model will assume that the mean of the response distribution is fixed to zero.
* there is now an option `bmm.silent` that allows to suppress messages
* the baseline activation `b` was removed from the `IMM` models, as this is internally fixed 
to zero for scaling and as of now cannot be predicted by independent variables because the model would be unidentifiable.
* the arguments used to fit the bmm model are now accessible in the `bmmfit` object via the `fit$bmm_fit_args` list.
* add class('bmmfit') to the object returned from fit_model() allowing for more flexible postprocessing of the underlying `brmsfit` object. The object is now of class('bmmfit', 'brmsfit')

### Bug Fixes
* an error with the treatment of distances in the `IMMfull` and the `IMMbsc` has been corrected. This versions ensures that only positive distances can be passed to any of the two models.
* removed a warning regarding the scaling of the distances in the `IMMfull` and the `IMMbsc` that was specific only for circular distances.

### Documentation
* All vignettes have been update to the new `bmmformula` syntax.


R/helpers-formula.R
```diff
@@ -45,9 +45,55 @@ get_response <- function(formula) {
...
+#' of the model. In general, measurement models do not all have a `mu`
+#' parameter, therefore it is more straigthforward to explicitely predict each
+#' parameter of the model.
+#'
+#' For example, in the following `brms` formula for the drift diffusion model,
+#' the first line corresponds to the drift rate parameter, but this is not
+#' explicitely stated.
+#'
+#'
+#'           brmsformula(rt | dec(response) ~ condition + (condition | id),
```
venpopov, line 75: I know we don't have a ddm model yet in bmm, but I think this is a nice example, because it indeed doesn't have a mu parameter. We can update this once we have a ddm implemented

R/helpers-formula.R
```diff
@@ -101,15 +101,15 @@ get_response <- function(formula) {
...
+#' imm_formula <- bmmformula(
 #'   c ~ 0 + setsize + (0 + setsize | id),
 #'   a ~ 1,
 #'   kappa ~ 0 + setsize + (0 + setsize | id)
 #' )
 #'
 #' imm_formula
 #'
-bmm_formula <- function(formula, ...){
+bmmformula <- function(formula, ...){
```
venpopov, line 112: I thing it would be good to have the same function naming style as brmsformula - otherwise it might be confusing for people who are used to brmsformula(), and forget to put underscore (e.g. me!)

R/bmmformula.R
```diff
@@ -0,0 +1,102 @@
...
+  # label the different formulas according to the parameter predicted
+  names(formula) <- pform_names
+  return(formula)
+}
+
+
+# alias of bmmformula
+#' @export
+bmf <- function(formula, ...) {
+  brmsformula(bmmformula(formula, ...))
```
GidonFrischkorn, line 101: Are you sure, that we have to pass this to `brmsformula`. Intuitively, this would convert the bmmformula to a brmsformula. But this should be done inside the configure_model.method, correct?

venpopov (approved):
Ok, done! Awesome job with the foundational code for the transition, now everything is also generalized and cleaned up. You can review and see if you have any questions/comments/suggestions. Let's not merge it immediately - I'd like to add a few unit tests for smaller functions, and update some of the documentation, before I bumpt the version number and we merge this. I will work on those while you review this

GidonFrischkorn:
Perfect, I will pull the current version of the branch and check out all the changes. I agree that we should not rush with updating this to avoid unnecessary patches due to some things we might have overlooked.

venpopov:
And as I mentioned on teams, here's the testing report, comparing model fits with the current branch version to the last release:

https://github.com/venpopov/bmm/blob/79-DV-passed-to-model/tests/internal/internal-model-tests.md

GidonFrischkorn:
The vignettes are now all updated to the `bmmformula` syntax. I have run all checks and found no errors.

venpopov:
> The vignettes are now all updated to the `bmmformula` syntax. I have run all checks and found no errors.

ok I'll look them over when I finish the hotfix for the setsize character/factor issue

GidonFrischkorn:
I have now also gone through all the changed files listed here on GitHub and had no further comments beyond the small commits that I made for the `IMMspatial` models. Is there anything more that I should take care of. Are there any tests that you think would be helpful to run.

venpopov:
I fixed the setsize checks. I want to ad a few more tests and finish documentation of new functions, which I'll do Saturday. You could look into updating the examples for all functions. Can you also describe the main changes and motivation in the main message for the pull request for the record, and update the news.md with the same info?

GidonFrischkorn:
> I fixed the setsize checks. I want to ad a few more tests and finish documentation of new functions, which I'll do Saturday. You could look into updating the examples for all functions. Can you also describe the main changes and motivation in the main message for the pull request for the record, and update the news.md with the same info?

Yes, definetely.

Actually, I will have a look into saving the examples in a separate _Examples_ folder and come up with a unified naming scheme, if you are ok with this. I felt that this might be easier than having the R-Code in the commented format. This way it would also be easier to check that all the examples are running with the current version of the package. I assume there should be an option in the checks to verify that there are no errors in the examples.

venpopov:
Sure, though does that still work with {don't run}? We don't want to run all models every time we build the package locally with check()

R/bmm_model_IMM.R
venpopov, line 1: I've been thinking for a while, and now is a good time to bring it up - to me spaPos is unintuitive as an argument. Without reading the documentation it sounds like it asks for the position of non_targets, whereas it asks for the distance. 
Can I suggest renaming this to spa_dist? More generally, both non_targets and spaPos refer to different features of the non_targets relative to the target. So mayb we should consider something like nt_features/nt_values and nt_dists. That is, I'm suggesting renaming:

non_targets to nt_features or nt_values
spaPos to nt_dists

makes it consistent with the parameter naming (thetant), and makes it clear what type of information each contains and how they are related. What do you think?

GidonFrischkorn: Agreed. I would suggest to rename 

`non_targets` to `nt_features`, and
`spaDist` to `nt_distance`

Generally, I think having the labels be as intuitive as possible and avoid abbreviations as much as possible.

I can add another commit that will include this change. For consistency, we should not only change it in the user facing function, but I would also try to change it in the specification of the `mixture3p` and the `IMM` models.

NEWS.md
```diff
@@ -1,3 +1,20 @@
+# bmm 0.3.0
+
+### New features
+
+* BREAKING CHANGE: The `fit_model` function now requires a `bmmformula` to be passed. The  syntax of the `bmmformula` or its short form `bmf` is equal to specifying a `brmsformula`. However, as of this version the `bmmformula` only specifies how parameters of a `bmmmodel` change across experimental conditions or continuous predictors. The response variables that the model is fit to now have to be specified when the model is defined using `model = bmmmodel()`. (#79)
+* BREAKING CHANGE: The `non_target` and `spaPos` variables for the `mixture3p` and `IMM` models were relabled to `nt_features` and `nt_distance` for consistency. This is also to communicate that distance is not limited to spatial distance but distances on any feature dimensions of the retrieval cues. Currently, still only a single generalization gradient for the cue features is possible.
```
venpopov, line 6: let's make them both plural for consistency - especially since both arguments take a vector of variable names for features or distances of each non_target variable. I'll make that change everywhere

NEWS.md
```diff
@@ -35,7 +52,7 @@
 
 ### New features
 
-* BREAKING CHANGE: Improve user interface to fit_model() ensures package stability and future development. Model specific arguments are now passed to the model functions as named arguments (e.g. `mixture3p(non_targets, setsize)`). This allows for a more flexible and intuitive way to specify model arguments. Passing model specific arguments directly to the `fit_model()` function is now deprecated (#43).
+* BREAKING CHANGE: Improve user interface to fit_model() ensures package stability and future development. Model specific arguments are now passed to the model functions as named arguments (e.g. `mixture3p(nt_features, setsize)`). This allows for a more flexible and intuitive way to specify model arguments. Passing model specific arguments directly to the `fit_model()` function is now deprecated (#43).
```
venpopov, line 55: since this release exists and can be installed, we should not rewrite history as it can be confusing. I'll change this back

tests/internal/generate_ref_model_fits.R
venpopov, line 1: this file uses version v0.2.1 for generating ref fits, so the syntax should not be changed to the new one (also the formula is still the old one so the change is inconsistent. I will revert it and then name the file with the bmm version for generating the ref fits. We can have different versions of this file for different bmm version ref fits

venpopov:
OK, this should be good to go. I will check the dev-notes and after that you can merge this

R/helpers-model.R
```diff
@@ -370,24 +397,53 @@ use_model_template <- function(model_name,
...
-                                  "   # save some variables as attributes of the data for later use (optional)\n\n\n",
+                                  "   # compute any necessary transformations (optional)\n\n",
+                                  "   # save some variables as attributes of the data for later use (optional)\n\n",
                                   "   data = NextMethod('check_data')\n\n",
                                   "   return(data)\n",
                                   "}\n\n",
                                   .open = "<<", .close = ">>")
 
+  # add bmf2bf method if necessary
+  bmf2bf_method <- glue::glue("#' @export\n",
```
venpopov, line 411: This comment is for the future, no need to do anything for this pull request. 

You know more about models that use more than one response variable. Once you are done with EZDM, can we generalize this function to have a bmf2bf.2resp? Or are there many different ways to specify 2 response variables depending on the type of model.

A second note is that there is code duplication from the other method. everything is the same except for the reponse part of the formula. So once you have the ezdm merged, we can see what can be refactored.

Issue #72, opened 2024-02-08
add check to SDM whether data is sorted by design matrix


venpopov:
to ensure that the adaptive denom calculation procedure is optimally calculated.

venpopov:
G.

venpopov:
Reopenining this to remind myself that I want to add the following message to the query about whether the user wants their data to be sorted:

```
To disable this check and query, use options("bmm.sort_data" = TRUE) to always sort or options("bmm.sort_data" = FALSE) to never check.
```

venpopov:
Also, while adding the above I noticed that the check did not actually run because when we added the global options, the default was no longer NULL but "check", and the is.null(sort_data) did not get executed.

PR #93, opened 2024-02-12
Feature/issue 72 add check to sdm whether data is sorted by design matrix

venpopov:
#### Summary

The sdm is much faster when the data is sorted by the design matrix. This pull request adds a check for the sdmSimple model whether the data is sorted by the design matrix. If not, it gives the user 3 options:
- to quit and sort the data themselves (provides them with code)
- to let bmm sort the data
- to continue with slower estimation

users can control the default behavior via new argument sort_data to fit_model, which can also be set as a global option:
- Null (default): the model asks the user the default query if it detects that the data is unsorted
- True: always sort the data
- False: do not check and do not ask

This check is in a general function order_data_query(), which can be used by any model that would benefit from sorted data.

#### Tests

#### Release notes

* add a check for the sdmSimple model if the data is sorted by predictors. This leads to much faster sampling. The user can control the default behavior with the `sort_data` argument (#72)


R/utils.R
```diff
@@ -78,14 +78,23 @@ softmaxinv <- function(p, lambda = 1) {
...
-      opts$chains <-  parallel::detectCores()
+      opts$chains <- parallel::detectCores()
     }
+  } else {
+    cores = NULL
   }
-  exclude <- c("parallel")
+  withr::local_options(
+    list(
+      mc.cores = parallel::detectCores(),
```
GidonFrischkorn, line 90: mc.cores = cores

venpopov:
This is ready to go

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: bmm
 Title: Easy and Accesible Bayesian Measurement Models using 'brms'
-Version: 0.3.0.9000
+Version: 0.3.1.9000
```
GidonFrischkorn, line 3: Should we remove the x.x.x.9000 at some point? Or have you left it in there for a specific purpose?

R/helpers-data.R
```diff
@@ -292,4 +293,31 @@ get_standata <- function(formula, data, model, prior=NULL, ...) {
 }
 
 
+# check if the data is sorted by the predictors
+is_data_ordered <- function(data, formula) {
+  predictors <- rhs_vars(formula)
```
GidonFrischkorn, line 298: I think this needs an additional check if some of the `rhs_vars` appear as `lhs_vars` indicating they are part of a non-linear formula. Probably, you can recycle some code from the `bmf2bf` function here. Would be nice if there is a solution without looping over all formulas.

But something like `dpars <- lhs_vars(formula)` and then `predictors <- predictors[!predictors %in% dpars]` should probably work.

GidonFrischkorn (commented):
To accomodate non-linear formulas in the `bmmformula` you probably need to adapt the extraction of the left-hand side arguments and evaluate if some of them also appear as right hand side arguments in the `bmmformula`. I have left a comment for you to check.

DESCRIPTION
venpopov: on the master it doesn't have it. on develop it does. this makes sure that when we make changes to develop, the main website is not updated - rather anything on develop shows up under https://venpopov.github.io/bmm/dev

That way the website is always up to date with the latest release, but we can also see the documentation and changes to the website relevant for the development version

R/helpers-data.R
venpopov: right, because the predictor then would not be present in the data. good catch

venpopov: ok, should be fixed now

GidonFrischkorn (approved):

PR #104, opened 2024-02-18
add log_lik and posterior_predict methods for sdmSImple

venpopov:
#### Summary

Add log_lik and posterior_predict methods for sdm simple
- turns out posterior_epred is not necessary - all the functionality that supposedly depends on it works without it

Now pp_check(), conditional_effects and bayes_factor from bridge sampling work

#### Tests

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bmm
 Title: Easy and Accesible Bayesian Measurement Models using 'brms'
-Version: 0.3.1.9000
+Version: 0.3.2.9000
 Authors@R: c(
     person("Vencislav", "Popov", , "vencislav.popov@gmail.com", role = c("aut", "cre", "cph")),
     person("Gidon", "Frischkorn", , "gidon.frischkorn@psychologie.uzh.ch", role = c("aut", "cph")))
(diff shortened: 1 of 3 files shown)


venpopov:
If Lea asks, I'm working on the lecture 😆

GidonFrischkorn:
So, I quickly tested the functionality and everything works. But I noticed that using the `brms::bayes_factor` function will recompile the models using `rstan`. Maybe we should add a note for users that should they be using `cmdstanr` for estimating the models, then Bayes Factor estimation will require a recompilation of the model in `rstan`. I do not think that this is problematic, so I will merge the changes and we can add the warning in a patch.

image

Btw, fit2 was a model fixing kappa over all set_sizes in the OberauerLin _2017 data. But for speed reasons I tested it without random effects.

GidonFrischkorn (approved):

venpopov:
For the bayes_factor on the sdm, it makes a dramatic difference whether you use priors. Right now the bayes_factor for the defaul behavior on kappa is not accurate. This is because if kappa only has an intercept you get the brms prior of Normal(5,0.8). Whereas when you include setsize with supressed intercept you get a flat prior on kappa. Which is why I started working on implementing the flexible default prior setting, so that this is not an issue.

PR #110, opened 2024-02-21
107 estimation of bayes factors for models including random effects requires brmssave pars

GidonFrischkorn:
#### Summary
- add function `save_pars_bmm2brms` to the package.
- add test to evaluate that the function works as intended

I wanted to avoid that there is a conflict if users have loaded both `brms` and `bmm` and therefore named the function differently. Generally, we might want to clarify that we recommend that users load both `brms` and `bmm` so that they can access the post-processing of the fit-objects that is supported by `brms`.
 
#### Tests


venpopov:
Thanks for doing this. I see a couple of problems around the current implementation. If people didn't use save_pars= argument and attempt bridge sampling they will get the error:

    stop2(
      "Bridgesampling failed. Perhaps you did not set ",
      "'save_pars = save_pars(all = TRUE)' when fitting your model? ",)

which then doesn't math the function name in bmm, so they will still have to load brms. We could add internal checks and rewrite this, but why? Also, if save_pars changes arguments in the future, we'll have to change our function too, which we might not knot until someone reports an error.

A much simpler way is like this:

```
#' @inherit brms::save_pars title params return
#' @description Thin wrapper around [brms::save_pars()]. When calling
#'   [fit_model] with additional information to save parameters you can use this
#'   function to pass information about saving parameter draws to `brms` without
#'   having to load `brms`. Alternatively, you can also load `brms` and call
#'   `save_pars`. For details see ?brms::save_pars.
#' @export
save_pars <- brms::save_pars
```

I implemented this and pushed the changes

venpopov (approved):
Thanks for doing this. I see a couple of problems around the current implementation. If people didn't use save_pars= argument and attempt bridge sampling they will get the error:

    stop2(
      "Bridgesampling failed. Perhaps you did not set ",
      "'save_pars = save_pars(all = TRUE)' when fitting your model? ",)

which then doesn't math the function name in bmm, so they will still have to load brms. We could add internal checks and rewrite this, but why? Also, if save_pars changes arguments in the future, we'll have to change our function too, which we might not knot until someone reports an error.

A much simpler way is like this:

```
#' @inherit brms::save_pars title params return
#' @description Thin wrapper around [brms::save_pars()]. When calling
#'   [fit_model] with additional information to save parameters you can use this
#'   function to pass information about saving parameter draws to `brms` without
#'   having to load `brms`. Alternatively, you can also load `brms` and call
#'   `save_pars`. For details see ?brms::save_pars.
#' @export
save_pars <- brms::save_pars
```

I implemented this and pushed the changes

PR #113, opened 2024-02-21
Feature/108 implement sdt models

GidonFrischkorn:
#### Summary
- Add SDT models for old/new (signal/noise) recognition as a `bmmmodel` class.
- Add distribution functions for density, and random generation for SDT models
- Initiate aggregate data method (experimental) for aggregating data to speed up model

#### Tests


GidonFrischkorn:
I just wanted to share my progress so far. The 2-AFC SDT models should be implemented. For now, I added all noise distributions that can be easily accommodated (Normal, EVD, Cauchy, and Logistic). However, we might want to discuss if there that matches the theoretical discussions around SDT models.

venpopov:
ok, I'll try to review what you have so far before we meet next week

GidonFrischkorn:
> ok, I'll try to review what you have so far before we meet next week

There is no hurry. I just thought you might be interested. As the 2-AFC models are mostly done. If you other things to do, I can also show you what I have done next week when we meet.

venpopov:
yes, I just want to consider more broadly how the various different SDT models such as old/new,2-AFC, m-AFC, confidence ratings might fit together and what we would need to have as separate models

venpopov:
> I found this article that should provide a good reference how to generalize the old/new SDT models to m-AFC SDT models: https://www.sciencedirect.com/science/article/pii/S0022249612000260

good find! Given his interest and extensive reasearch and writing on SDT, SDT via regression and bayesian implementations, measurement in general (https://www.tc.columbia.edu/faculty/ld208/), maybe at some point we can contact him and chat about the package and get his thoughts on how the SDT is implemented, maybe get him involved if he wants to?

GidonFrischkorn:
I thought the same. For the moment, I think we can implement plenty of things ourselves, but it would still be interesting to discuss with him about optimizations with respect to the SDT implementations.

He has also done some interesting work linking SDT to psychometric models such as IRT that I would find interesting to explore. But then again, there is already plenty of work on my desk. So maybe this is rather something for the future...

R/bmm_model_SDT.R
```diff
@@ -0,0 +1,182 @@
+#############################################################################!
+# MODELS                                                                 ####
+#############################################################################!
+# see file 'R/bmm_model_mixture3p.R' for an example
+
+.model_SDT <- function(response, stimulus, nTrials = NULL, dist_noise = "gumbel", ...) {
+   out <- list(
+      resp_vars = nlist(response, stimulus, nTrials),
```
venpopov, line 8: shouldn't stimulus and nTrials be under "other_vars"?

R/bmm_model_SDT.R
```diff
@@ -0,0 +1,182 @@
...
+      other_vars = nlist(dist_noise, ...),
+      info = list(
+         domain = 'Perception & Recognition',
```
venpopov, line 11: we should discuss a nomenclature for these more generally

R/bmm_model_SDT.R
```diff
@@ -0,0 +1,182 @@
...
+# See ?bmf2bf for details.
+# (YOU CAN DELETE THIS SECTION IF YOUR MODEL USES A STANDARD FORMULA WITH 1 RESPONSE VARIABLE)
+
+#' @export
+bmf2bf.SDT <- function(model, formula) {
+   # retrieve required response arguments
+   response <- model$resp_vars$response
+   stimulus <- model$resp_vars$stimulus
+   if (!is.null(model$resp_vars$nTrials)) {
+      nTrials <- model$resp_vars$nTrials
```
venpopov, line 109: this can be moved in the same if clause below rather than having two separate ones

R/bmm_model_SDT.R
```diff
@@ -0,0 +1,180 @@
+#############################################################################!
+# MODELS                                                                 ####
+#############################################################################!
+# see file 'R/bmm_model_mixture3p.R' for an example
+
+.model_SDT <- function(response, stimulus, nTrials = NULL, dist_noise = "gumbel", ...) {
```
venpopov, line 6: default should probably be the normal as that is the most common :)

R/distributions.R
```diff
@@ -600,3 +600,164 @@ rIMM <- function(n, mu=c(0,2,-1.5), dist = c(0,0.5,2),
...
+#' @param x A vector of observed responses.
+#' @param p A vector of probabilities.
+#' @param q A vector of quantiles.
+#' @param n Number of observations to generate data for.
+#' @param size The number of trials used to collect the responses.
+#' @param dprime A vector of numerical values for the signal strength.
+#' @param crit A vector of numerical values for the decision criterion.
+#' @param stimulus A character vector ("signal"/"noise", "old"/"new") or
+#'   integer vector (1/0) coding which stimuli were shown alongside the observed responses,
+#'   or for which to generate data for.  If no information is provided for `rSDT` then we will
```
venpopov, line 621: the function only returns hits by default. I think this is probably for the best, and would rather remove the description than change the function. Or return a list of two vectors, one for hits and one for FAs

R/distributions.R
```diff
@@ -600,3 +600,164 @@ rIMM <- function(n, mu=c(0,2,-1.5), dist = c(0,0.5,2),
...
+    probs <- stats::plogis(acts)
+  }
+
+  quantile <- stats::qbinom(p = p, size = size, prob = probs, log.p = log.p)
+  quantile
+}
+
+#' @rdname SDTdist
+#' @export
+rSDT <- function(n, size, dprime = 1, crit = dprime/2, stimulus = 1, dist_noise = "normal") {
```
venpopov, line 736: could be good to have an option to return proportions instead of counts

R/distributions.R
```diff
@@ -600,3 +600,164 @@ rIMM <- function(n, mu=c(0,2,-1.5), dist = c(0,0.5,2),
...
+    probs <- stats::pnorm(acts)
+  } else if (tolower(dist_noise) == "gumbel") {
+    probs <- evd::pgumbel(-acts, lower.tail = FALSE)
+  } else if (tolower(dist_noise) == "cauchy") {
+    probs <- stats::pcauchy(acts)
+  } else if (tolower(dist_noise) == "logistic") {
+    probs <- stats::plogis(acts)
+  }
+
+  density <- stats::dbinom(x, size = size, prob = probs, log = log)
```
venpopov, line 663: this should be log=TRUE

R/bmm_model_SDT.R
```diff
@@ -0,0 +1,185 @@
...
+   # retrieve required response arguments
+   response <- model$resp_vars$response
+   stimulus <- model$resp_vars$stimulus
+
+   # set the base brmsformula given the variable names
+   if (!is.null(model$resp_vars$nTrials)) {
+      nTrials <- model$resp_vars$nTrials
+      brms_formula <- brms::bf(paste0(response," | ", "trials(",nTrials,")", " ~ dprime*",stimulus," - crit" ), nl = TRUE)
+   } else {
+      brms_formula <- brms::bf(paste0(response," ~ dprime*",stimulus," - crit"),nl = TRUE)
```
venpopov, line 119: this works for symmetric distributions, but for assymetric ones it gives the wrong part of the cdf. Will explain in person

R/distributions.R
```diff
@@ -600,3 +600,164 @@ rIMM <- function(n, mu=c(0,2,-1.5), dist = c(0,0.5,2),
...
+#'   otherwise P\[X > x\].
+#'
+#' @keywords distribution
+#' @export
+#'
+#' @examples
+#' # example code
+#'
+#'
+dSDT <- function(x, size, dprime = 1, crit = dprime/2, stimulus = 1, dist_noise = "normal", log = FALSE) {
```
venpopov, line 638: all of these need a check if the dist_noise is one of the valid options

R/bmm_model_SDT.R
GidonFrischkorn: resolved by https://github.com/venpopov/bmm/pull/113/commits/f485e2a5bab8c1ad627a71e581e907bdd54ed570

R/bmm_model_SDT.R
GidonFrischkorn: Agreed. I will open an issue to discuss labels we want to use across models to categorize them consistently.

R/distributions.R
GidonFrischkorn: Maybe you can have a look if this resolves your concerns.

R/distributions.R
GidonFrischkorn: Hmm... do you mean that we return the proportion of hits/FA instead of the count. Why would users need these? If they want they can directly compute them from `returnedCounts/size`, but they would loose the information on how many trials the proportion was calculated on.

Maybe I am misunderstanding something, but my thought was to make these functions consistent with the model implementation.

R/bmm_model_SDT.R
GidonFrischkorn: As far as I have checked with the parameter recoveries I ran, this implementation does recover the parameters well using the `cloglog` link for Gumbel noise. But we should test this thoroughly to make sure that I did not miss anything.

GidonFrischkorn:
Things that need to be added for finishing up the Old/New Recognition SDT models.

- [ ] vignette introducing the model and providing an example how to fit the models
- [ ] unit tests for the added functions
- [ ] add an example data set to fit the models
- [ ] evaluate parameter recovery more systematically

GidonFrischkorn:
I will see that I take care of most of the to dos in the next week.

venpopov:
Perfect, and I can finish the review and test it with the code you sent in the meantime

GidonFrischkorn:
I just found this package implementing EV and UEV SDT models in STAN with a newly proposed link function that allows to fit hierarchical models also for the thresholds: https://github.com/boryspaulewicz/bhsdtr2/tree/master

Maybe we should have a look at how they coded this in STAN and see if we can adapt this for our package. Generally, the principle for predicting the model parameters looks pretty similar, with the difference that they do interact with brms at all.

GidonFrischkorn:
After some consideration I suggest we close this pull request, given that the codebase has changed considerably since I worked on this. In addition the models that have been implemented were not too complicated, so it should be fairly quick to reimplement them based on the current codebase.

Moreover, there are some major additions that I would like to include to accommodate current research.
1. https://osf.io/preprints/psyarxiv/qhrfj_v3 The recent publication on extreme value SDT
2. https://scholar.google.com/citations?view_op=view_citation&hl=en&user=6AXZPtwAAAAJ&sortby=pubdate&citation_for_view=6AXZPtwAAAAJ:08ZZubdj9fEC Work on generalized threshold functions that allow for a hierarchical estimation of thersholds without violating order constraints.

If you want to discuss these changes in more detail, I am happy to chat. I have made a document outlining the most relevant steps for implementing these and am happy to share my ideas.

venpopov:
Sounds good

Issue #102, opened 2024-02-17
allow specify nt_features and nt_distances with partial name rather than a vector?


venpopov:
Could be nice instead of having to fully write the vector of column names for non_target features and distances, to be able to just specify the prefix of columns, and we get them internally via "grepl". Logic in the data check would be:

- check if nt_features or nt_distances match existing column names - continue as usual
- if not, check if grepl matches existing column names - if so, get those column names and continue as usual
- if both return null, give the current error

not high priority, but nice to have

GidonFrischkorn:
I like this idea. In princinple, if users provide the character part of the variables, then we could infer it assuming that they would always start from 1 and then go-on until `max(setsize)-1`. We could allow to only provide a single character to be pased to `nt_features` and `nt_distances`, but still throw a warning that we assume that they used the above mentioned numbering scheme at the end of the provided character.

With some regex syntax the user might even specify something like `nt*_col` with the * reflecting the integers indicating the non-target identifier. But this would still, require some instructions, that we could probably add to the documentation.

venpopov:
we don't have to assume how the numbering goes. we can just extract all columns that contain the character part

I like the idea of the regex syntax also

PR #114, opened 2024-02-21
Feature/issue 102 allow specify nt features and nt distances with partial name rather than a vector

venpopov:
#### Summary

Allow users to specify model variable arguments such as nt_features and nt_distances via regular expressions.

- Transform check_model into an S3 method
- add regex argument to mixture3p and IMM models
- add atribute 'regex_vars' to the models, which specifies which variables can be specified via regular expressions
- the check_model.bmmmodel method calls a new function `replace_regex_variables(model, data)`. If regex is FALSE or NULL or MISSING, the function just returns the model object, making it backwards compatible with all models regardless of whether they have a regex argument or not. if regex=TRUE, the functions checks which variables can be specified via regular expressions, and then looks in the data column names for matching variables
- update vignettes and examples to document new option
- tests for new functionality
- fix some issues with IMM vignete's last figure mismatching parameters

#### Example usage:

```
data <- OberauerLin_2017
model <- IMMfull(resp_err = "dev_rad",
                 nt_features = 'col_nt',
                 nt_distances = 'dist_nt',
                 setsize = 'set_size',
                 regex = TRUE)
```

or

```
data <- OberauerLin_2017
model <- IMMfull(resp_err = "dev_rad",
                 nt_features = 'col_nt[1-7]',
                 nt_distances = 'dist_nt[1-7]',
                 setsize = 'set_size',
                 regex = TRUE)
```

or 

```
data <- data.frame(item1_color = runif(10,-pi,pi),
                   item2_color = runif(10,-pi,pi),
                   item3_color = runif(10,-pi,pi),
                   dev_rad = rnorm(10,0,0.2))

model <- mixture3p(resp_err = 'dev_rad',
                   nt_features = 'item.*_color',
                   setsize = 4,
                   regex = T)
```

#### Tests

#### Release notes
* you can now specify variables for models via regular expressions rather than character vectors [#102]


GidonFrischkorn (commented):
Generally, the changes look good and I got everything to run on my machine, too. However, I noticed that using the `regex` option can also use problems, for example if I were to fit the model only to `setsize = 4` in the `OberauerLin_2017` data then I would have to delete all `col_nt` and `dist_nt` vars for the nts 4 to 7. However with the `paste0("col_nt",1:3)` this does not cause problems.

Maybe we can add this when introducing the `regex` option. That this will only work if all variables following a certain naming scheme are used for the model specification. Otherwise, in the above mentioned example, I get an error that there are more `nt_features` than `max(setsize) - 1` 

I think this is not too big of an issue, but users might still be confused about it when trying to run models only on specific conditions without adapting their dataset.

venpopov:
> Generally, the changes look good and I got everything to run on my machine, too. However, I noticed that using the `regex` option can also use problems, for example if I were to fit the model only to `setsize = 4` in the `OberauerLin_2017` data then I would have to delete all `col_nt` and `dist_nt` vars for the nts 4 to 7. However with the `paste0("col_nt",1:3)` this does not cause problems.
> 
> Maybe we can add this when introducing the `regex` option. That this will only work if all variables following a certain naming scheme are used for the model specification. Otherwise, in the above mentioned example, I get an error that there are more `nt_features` than `max(setsize) - 1`
> 
> I think this is not too big of an issue, but users might still be confused about it when trying to run models only on specific conditions without adapting their dataset.

I thought about that too, but this is not a problem if you specify the regex with a limited number range. E.g., this will happen if you call

```
mixture3p(resp_err = 'dev_rad', nt_features = 'col_nt', setsize = 4, regex=T)
```

but not if you call

```
mixture3p(resp_err = 'dev_rad', nt_features = 'col_nt[1-3]', setsize = 4, regex=T)
```

This example is included in the vignettes. In general, we give the regex option, but it's not our task to educate people about regex - it is there to use for those who want it and know how to use (as in any other package that offers regex arguments, for example a few functions in brms, posterior and tidybayes)

GidonFrischkorn:
I see your points and agree that both the cases I highlighted will likely be rare. So, let's merge this for now and if additional problems pop up the more we use the feature we can re-evaluate if we need to change something about this.

GidonFrischkorn (approved):

PR #119, opened 2024-02-22
restructure call to model objects for missing args

venpopov:
#### Summary

Adds NULL default arguments to .model_* functions and changes alias of model functions to handle missing arguments. This handles missing required arguments more gracefully, allowing for easier testing and documentation (not needing to call IMMfull(NA, NA, NA, NA) if you just want to test it.

#### Tests


GidonFrischkorn (commented):
This looks good. However, I wondered if we should limit the `NULL` default argument only to the `other_vars` so that models cannot be specified without passing `resp_vars`. I know that this partially defeats the purpose of simplifying the documentation. But at least, it would reduce the number of `NA` calls required in the automatically generated documentation.

What do you think?

venpopov:
The way it's implemented now when users call any model, they have to specify all variables. The default NULL arguments are only for the internal function. The user-facing alias is now updated to, for example:

```
IMMfull <- function(resp_err, nt_features, nt_distances, setsize, ...) {
  stop_missing_args()
  .model_IMMfull(resp_err = resp_err, nt_features = nt_features,
                 nt_distances = nt_distances, setsize = setsize, ...)
}
```
This will give an error if any of the arguments that don't have a default NULL are not specified. It recognizes that if any arguments have a default that's different from NULL (e.g. in regex in the other pull request), users don't have to specify it. So now this handles both cases - users have to specify the variables, but we can use the internal functions in documentation and tests without.

venpopov:
Weird, the vignettes fail to build. Let me look into this before we merge

venpopov:
Ah, I need to accomodate the new regex argument from the other PR into the new aliases

GidonFrischkorn:
I am already on it.

GidonFrischkorn:
I made a mistake when resolving the conflicts between the develop branch and this one.

GidonFrischkorn (approved):

PR #121, opened 2024-02-22
Feature/issue 118 should we load brms automatically when bmm is loaded

venpopov:
#### Summary

Switch brms from Imports to Depends to load it when loading bmm
Remove save_pars alias as its no longer necessary

#### Tests


GidonFrischkorn (approved):
Looks good. And checking these minor changes on the phone works actually pretty well.

Can you bump the version number for this patch? Or do you think this is not necessary for this change.

venpopov:
I'll bump it. We don't have to always do it, but this is a change in behavior so makes sense

PR #122, opened 2024-02-22
Bigfix/issue 120 is data ordered fails with a complex formula

### LINKED ISSUE #120: is_data_ordered fails with a complex formula
Using the |p1| syntax for correlated random effects across parameter causes an error in the sdmSimple

```
formula <- bmf(c ~ 0 + set_size + (0 + set_size | p1 | id),
                   kappa ~ 0 + set_size + (0 + set_size | p1 | id))
```

image

venpopov:
#### Summary

Fixes #120. Also fixes it to work with non-linear predictors

#### Tests


GidonFrischkorn (approved):
Looks good. Given that the tests you coded passed, I do not think that I have to check it on my machine.

PR #124, opened 2024-02-23
Add update.bmmfit method

venpopov:
#### Summary

A new update.bmmfit method reverts postprocessing if necessary, does all the typical bmm checks and transformations, then passes them to update.brmsfit.

- now also stores more bmm stuff under fit$bmm, such as the model, the user provided bmf(), etc, things we might need to access for other postprocessing functions, such as the planned custom summary

#### Tests


GidonFrischkorn (approved):
Great work. Thanks for checking this out. And super nice that this is now also working for `bmmmodels`.

PR #137, opened 2024-02-27
Feature/117 add a message upon loading the package

GidonFrischkorn:
#### Summary
- add a package startup message.
- by convention this is added in the zzz.R file

I just thought that I share my implementation and then you can have a look at what is there and add things if there are settings that you think we should set `.onLoad`

#### Tests


venpopov:
perfect, will checkout now

venpopov (approved):
Looks good. I added a new commit:

- new function bmm_options and use it onLoad
- bmm_options() allows to view or set all bmm global options
- it is called onLoad to set defaults and display them in the package load message
- add tryCatch2 to be able to handle messages and display them when desired
- version bump and news entry

When you load the package now it looks like this:

image

PR #140, opened 2024-02-29
Transition to get_prior, make_stancode and make_standata

venpopov:
#### Summary

Since [Paul agreed](https://github.com/paul-buerkner/brms/pull/1604#issuecomment-1970005388) to make get_prior, make_stancode and make_standata into generic S3 methods, we can now replace get_model_prior, get_stancode and get_standata with:

- get_prior.bmmformula
- make_stancode.bmmformula
- make_standata.bmmformula

That way users of `bmm` can use the same functions get_prior, make_stancode and make_standata with bmm models just like they would do with `brms`.

This requires an installed version of brms > 2.20.14, so I have put in checks in the code to ensure that everything works with older versions as well.

Once the next release of brms is available on CRAN, we can make it a hard dependency and get rid of the old syntax completely.

#### Tests

* [X] Confirm that all tests passed
* [X] Confirm that devtools::check() produces no errors

- add new tests for the new methods, which are performed only if the installed version of brms is > 2.20.14


GidonFrischkorn (approved):
Thanks for taking care of this. Looks great and nice that we can keep the original functions as deprecated for now.

PR #143, opened 2024-03-03
Feature/issue 132 renaming get model prior

venpopov:
#### Summary

Some updates to the new methods for get_prior, make_stancode and make_standata. See [here](https://github.com/paul-buerkner/brms/pull/1604#issuecomment-1974044897) for the reasons.

#### Tests


GidonFrischkorn (approved):

GidonFrischkorn:
I tested it on my machine and did not run into any issues. Thanks for working this out together with Paul!

GidonFrischkorn:
The only thing I noticed. That despite updating `brms` I still got the warning that `get_model_prior` is deprecated. Although I used `default_prior` or `get_prior`.

venpopov:
ah, good catch, will fix it

PR #144, opened 2024-03-03
Summary method for bmmfit

### LINKED ISSUE #149: simplify structure of .model_*
Currently we have:

```
   out <- list(
      resp_vars = nlist(resp_err),
      other_vars = nlist(),
      info = list(
         domain = 'Visual working memory',
         task = 'Continuous reproduction',
         name = 'Signal Discrimination Model (SDM) by Oberauer (2023)',
         citation = paste0('Oberauer, K. (2023). Measurement models for visual working memory - ',
                           'A factorial model comparison. Psychological Review, 130(3), 841-852'),
         version = 'Simple (no non-targets)',
         requirements = '- The response variable should be in radians and represent the angular error relative to the target',
         parameters = list(
            mu = 'Location parameter of the SDM distribution (in radians; by default fixed internally to 0)',
            c = 'Memory strength parameter of the SDM distribution',
            kappa = 'Precision parameter of the SDM distribution (log scale)'
         ),
         links = list(mu = 'identity',
                      c = 'log',
                      kappa = 'log'),
         fixed_parameters = list(
            mu = 0
         )),
      void_mu = FALSE
   )
```

I propose we remove the "info" subset and unpack it as:

```
  out <- list(
    resp_vars = nlist(resp_err),
    other_vars = nlist(),
    domain = 'Visual working memory',
    task = 'Continuous reproduction',
    name = 'Signal Discrimination Model (SDM) by Oberauer (2023)',
    citation = paste0(
      'Oberauer, K. (2023). Measurement models for visual working memory - ',
      'A factorial model comparison. Psychological Review, 130(3), 841-852'
    ),
    version = 'Simple (no non-targets)',
    requirements = '- The response variable should be in radians and represent the angular error relative to the target',
    parameters = list(mu = 'Location parameter of the SDM distribution (in radians; by default fixed internally to 0)',
                      c = 'Memory strength parameter of the SDM distribution',
                      kappa = 'Precision parameter of the SDM distribution (log scale)'),
    links = list(mu = 'identity',
                 c = 'log',
                 kappa = 'log'),
    fixed_parameters = list(mu = 0),
    void_mu = FALSE
  )
```

it will simplify a bunch of the code when accessing elements of a fitted object (i just had to write fit$bmm$model$info$parameters and got annoyed 

thoughts @GidonFrischkorn ?

venpopov:
## Summary

The custom summary method is almost done. Below are some examples. There are two things left to do:

- how to determine the links for parameters

As you can see below, I have not yet implemented the links summary. It's tricky. For the sdm, I can just use the formula. But for the remaining models, because they use non-linear formulas, the link is not actually explicit in the formula - there are links only for the kappa1, kappa2, etc. One idea is to add an argument to the model$info, e.g., model$info$links, where we just specify the links even if we use that only for printing. Thoughts

- deal with the case where participants predict a fixed parameter (such as mu or mu1)

This is not difficult, but I will do that after we figure out the links as its less important.

(the colors can be turned off with options(bmm.color_summary = FALSE))

### sdm BEFORE
image

### sdm AFTER
image

### 2p BEFORE
image

### 2p AFTER
image

### 3p BEFORE
image

### 3p AFTER
image

### IMMfull before
image

### IMMfull after
image

Also closes #149 

## Tests

[] Confirm that all tests passed
[] Confirm that devtools::check() produces no errors

#### Release notes


GidonFrischkorn:
This looks great. I like the color and the highlighted labels. There are some typos (at least in the "Additional Fixed Constant Parameters" label). And I think this label is somewhat redundant, I would propose to just say either `Fixed Parameters` or `Constant Parameters`.

With respect to the links: I think adding them to the model info makes sense. In fact, for the `M3custom` we already have an additional argument that sets the links. With adding them to the model info, we could think about enabling to set links a more general feature where reasonable. This might not make sense for most of the models and compromise the stability of estimation, but in some cases users might prefer using different links.

venpopov:
ok, then let's meet one of the following days and you can show me again how you set the links in M3custom, so we can figure out how to set it for all mdoels before I implement that part of the summary method

GidonFrischkorn:
1. Add links to model info. Either hard coded for models that should only use a specific set of links. Or add `links` argument to the model call if users should be able to choose which links should be used for the different parameters.

3. Select parameters to be reported by checking if there are non-linear transformations in the model formula, and report all model parameters without non-linear transformations, and non-linear parameters if they were used for predicting a model parmeter.

venpopov:
This is now done. Took me much longer than I expected, but now all features should be ironed out. Here's the output for a few models:

image

image

image

image

image

image

image

There is an option "backend" for the summary method, with which you can always see the brms-type summary:

```r
> summary(fit, backend='brms')
 Family: mixture(von_mises, von_mises, von_mises, von_mises, von_mises) 
  Links: mu1 = identity; kappa1 = log; mu2 = identity; kappa2 = log; mu3 = identity; kappa3 = log; mu4 = identity; kappa4 = log; mu5 = identity; kappa5 = log; theta1 = identity; theta2 = identity; theta3 = identity; theta4 = identity; theta5 = identity 
Formula: dev_rad ~ 1 
         mu1 ~ 1
         kappa ~ 0 + SetSize
         a ~ 0 + SetSize
         c ~ 0 + SetSize
         kappa5 ~ 1
         mu5 ~ 1
         theta1 ~ c + a
         kappa1 ~ kappa
         kappa2 ~ kappa
         theta2 ~ LureIdx1 * (a) + (1 - LureIdx1) * (-100)
         mu2 ~ Item2_Col_rad
         kappa3 ~ kappa
         theta3 ~ LureIdx2 * (a) + (1 - LureIdx2) * (-100)
         mu3 ~ Item3_Col_rad
         kappa4 ~ kappa
         theta4 ~ LureIdx3 * (a) + (1 - LureIdx3) * (-100)
         mu4 ~ Item4_Col_rad
   Data: data (Number of observations: 4000) 
  Draws: 4 chains, each with iter = 500; warmup = 250; thin = 1;
         total post-warmup draws = 1000

Regression Coefficients:
                 Estimate Est.Error l-95% CI u-95% CI Rhat Bulk_ESS Tail_ESS
mu1_Intercept        0.00      0.00     0.00     0.00   NA       NA       NA
mu5_Intercept        0.00      0.00     0.00     0.00   NA       NA       NA
kappa5_Intercept  -100.00      0.00  -100.00  -100.00   NA       NA       NA
kappa_SetSize1       3.07      0.05     2.97     3.16 1.01     1689      658
kappa_SetSize2       2.45      0.05     2.35     2.55 1.00     1643      785
kappa_SetSize3       2.14      0.06     2.02     2.26 1.00     1400      753
kappa_SetSize4       2.01      0.09     1.83     2.17 1.00     1361      837
c_SetSize1           3.95      0.27     3.47     4.50 1.01     1028      613
c_SetSize2           3.38      0.24     2.94     3.87 1.00     1039      665
c_SetSize3           2.84      0.17     2.52     3.20 1.00      829      487
c_SetSize4           2.70      0.18     2.39     3.08 1.00      950      638
a_SetSize1           0.00      0.00     0.00     0.00   NA       NA       NA
a_SetSize2          -0.43      0.34    -1.11     0.25 1.00     1041      709
a_SetSize3          -0.90      0.28    -1.45    -0.38 1.00      921      594
a_SetSize4          -1.66      0.27    -2.20    -1.14 1.00      941      603

Draws were sampled using sample(hmc). For each parameter, Bulk_ESS
and Tail_ESS are effective sample size measures, and Rhat is the potential
scale reduction factor on split chains (at convergence, Rhat = 1).
```

R/bmmformula.R
venpopov, line 1: I combined the files bmmformula and helpers-formula, since we don't need two for that and it wasn't cclear what should go where

R/bmmformula.R
```diff
@@ -131,39 +133,219 @@ bmf <- function(formula, ...) {
...
+rhs_vars.formula <- function(formula, ...) {
   all_vars <- all.vars(formula)
   all_vars[-1]
 }
 
 
-is.formula <- function(x) {
+# adds an attribute nl to each component of the the formula indicating if the
+# any of the predictors of the component are also predicted in another component
+assign_nl <- function(formula) {
```
venpopov, line 314: this function is now used in constructing bmmformula, so for the future we can always just check which parameters are non-linear via is_nl() functionbellow, rather than always checking manually if the predictors are also parameters

R/restructure.R
venpopov, line 1: an initial restructure function allows us to apply new methods to fits using older bmm versions. there are more things to be added but for now it works for the summary function at least

R/summary.R
```diff
@@ -0,0 +1,239 @@
+#' Create a summary of a fitted model represented by a \code{bmmfit} object
+#'
+#' @inheritParams brms::summary.brmsfit
+#' @param backend Choose whether to display the *bmm* summary method (default),
+#'   or to display the *brms* summary method.
+#' @seealso \code{\link{summary.brmsfit}}
+#' @note You can turn off the color output by setting the option
+#' options(bmm.color_summary = FALSE) or bmm_options(color_summary = FALSE)
+#' @export
```
venpopov, line 9: summary.bmmfit first calls the summary.brms method, which extracts all necessary info for printing. then it passes it to print.bmmsummary

R/summary.R
```diff
@@ -0,0 +1,239 @@
...
+rename_mu_smry <- function(x, mu_pars) {
+  for (i in seq_along(x)) {
+    if (is.data.frame(x[[i]])) {
+      rownames(x[[i]])[rownames(x[[i]]) %in% mu_pars] <- paste0("mu_", mu_pars)
+    }
+  }
+  x
+}
+
+select_pars <- function(x) {
```
venpopov, line 144: selects which parameters to include in the summary - model parameters and any parameters used in non-linear formulas. because brms does not print out non-linear parameters, we don't either. E.g., if you predict kappa by a non-linear formula, kappa won't be included in the output, but just the parameters by which it is predicted.

GidonFrischkorn:
Thanks for implementing the additional changes! I am in a workshop tomorrow and will have to see when I get to review this throughly.

But if I split it up, I might get to it this evening and tomorrow.

GidonFrischkorn:
Another thing that just came to my mind is that we need to update the model template accordingly. If you already did so, then just ignore this comment.

venpopov:
> Another thing that just came to my mind is that we need to update the model template accordingly. If you already did so, then just ignore this comment.

not yet. I also just realized the vignettes need to be updated. 

I will update the vignettes and include that here.

For the model template, I suggest we wait. We also need to update the dev notes as well. My suggestion is to open an issue to update those, and we do it before the next release, as we might make more changes that affect them

venpopov:
I updated the vignettes. 

Also I added a fille bmmexamples in tests/local, which creates several internal datasets:

- bmmfit_example1, bmmfit_sdm_vignette, bmmfit_mixture2p_vignette, bmmfit_imm_vignette

these are stored inside a R/sysdata.rda file and are accessible internally like bmm:::bmmfit_imm_vignette

I reduced the size of each and now use these in the vignettes. Installed size is now down from 32mb to 5.5mb, but still get a note.

GidonFrischkorn (approved):
As far as I checked, this looks great. I even merged this into the latest dev version for the M3 implementation and the summary also works for the custom models with users naming the parameters they use in the activation formulae.

GidonFrischkorn:
Btw, locally on my Mac all checks are fine. I do not know what might be the problem regarding the R-CMD-checks performed by GitHub...

venpopov:
I figured it out, it's because I generated the example bmm fits with cmdstanr, and use one of them in the tests. Turns out that the object "remembers" it was made with cmdstanr and gives a warning that the package doesn't exist in the online check...

GidonFrischkorn:
Great that you found the problem! From my side we should be good to merge then.

PR #151, opened 2024-03-08
Restructure function and internal clean-up

### LINKED ISSUE #147: make restructure() function
as part of the new summary method, I realized it won't work for models fitted with older bmm versions (because of the links argument).  We can ensure it works by adding a restructure() function similar to the one in brms that transforms fits from older bmm functions to the current version. We can make it a restructure.bmmfit method after [this](https://github.com/paul-buerkner/brms/issues/1611)

an initial function is part of this commit from the summary branch: https://github.com/venpopov/bmm/blob/9ff528f61ed6e2e67ec760b4b919cfcc13c71c91/R/restructure.R

but it will need to be expanded and generalized

Tasks:

* [ ] clean-up fit$bmm stored information #148 
* [ ] clean-up stored environments with various objects
* [ ] s3 method for brms

### LINKED ISSUE #148: fit$bmm$config_args contains a lot of duplicated info
for ease we stored all the config args from fit_model to the bmmfit object, but a lot of it duplicates information stored elsewhere in the brmsfit object, so we should only keep what is unique, and see where we use this (I think in the update method), to just reconstruct it from the brmsfit object

venpopov:
#### Summary

Mostly internal cleanup

1. Simplifies the structure of *$bmm and cleans up environments stored with formulas. Before we were storing the entire environment unintentionally leading to larger objects that usual.

2. Finish the restructure() method. Now any model fitted after version 0.2.0 can be used with all new methods. For example, the fit of Klaus to Unsworth data now works with the new summary. As we change things in the future, we should add things to the restructure function when necessary

Closes #147. Closes #148 

#### Tests


GidonFrischkorn (approved):
Thanks for taking care of this cleanup. I did not find any issues. And the updates work fine on my machine.

PR #152, opened 2024-03-08
Allow to fix or predict any parameter

### LINKED ISSUE #72: add check to SDM whether data is sorted by design matrix

### LINKED ISSUE #142: allow users to fix any parameter of the model in the formula
e.g.

bmf(kappa = 3, thetat ~ 0 + setsize, thetant ~ 0 + setsize)

and we translate it to a constant prior

(it may already work by default actually, but Paul mentioned he doesn't like that and he might deprecate it)

### LINKED ISSUE #145: if predicting mu and you suppress the intercept, it gives an error
```r
data1 <- dplyr::filter(OberauerLin_2017, set_size %in% c(1,2,3,4), ID %in% 1:10)
formula <- bmf(c ~ set_size, kappa ~ 1, mu ~ 0 + set_size)
prior <- set_prior("normal(0,0.5)", class = "Intercept", dpar='mu')
fit3 <- fit_model(formula, data1, model,
                  parallel = T,
                  prior = prior,
                  backend = 'cmdstanr',
                  sort_data = T,
                  iter = 100)
```

```
Error: The following priors do not correspond to any model parameter: 
Intercept ~ constant(0)
```

if you don't supress the intercept, you can specify a prior and solve the issue. But if you suppress it, the constant prior remains. We should build a check before adding constant priors if the parameter is predicted. Related to #142

venpopov:
#### Summary

A few major changes in this PR:

- Predicting parameters fixed to a constant by default now works. You can just specify `bmf(mu ~ setsize)`, for example, for the sdmSimple() model, and similarly for mu1 for all the others
- Can now set any parameter to a constant by specifying `bmf(par = value)`. For example, if you want to fix kappa to 2, and not estimate it, you can use a formula like `bmf(c ~ 0 + setsize, kappa = 2)`

To achieve this, I made the following structural changes:

- `add_missing_parameters()` now tags each parameter in the bmmformula with an attribute constant. TRUE if it is fixed by default, and not predicted by the user, or if the user fixed a free parameter to a constant. Then it replaces all forms 'par = value' with 'par ~ 1'
- `check_model()` now updates the fixed_parameters field of the model object based on the "constant" attributes in the formula. If a default constant parameter is predicted in the user formula, it removes it from fixed_parameters. If a free parameter is fixed by the user, it adds it
- the default priors for all models are no longer specified in the configure_model.*() methods. These methods now return just the formula (with the family attached internally to it), the data, and any additional properties, but not the prior.
- the previous `prior_list` argument within configure_model(), that specifies the default prior is now specified in the model object as "default_priors".
- a new generic S3 method `configure_prior()` is called in fit_model() right after configure_model(). Like all other methods, it takes the model object, the data and the user formula, and in addition, the user given prior. The configure_prior.bmmmodel() method combines informaiton from the formula, data and fixed_parameters + default_priors field of the model object, and constructs both the constant priors and the default_priors. 
- the above is now much cleaner and reduces duplication of code. custom methods for configure_prior.modelname() need to be added only for special priors that cannot be constructed as above - for example, the prior for setsize 1 in the non-target mdoels

I also did a bunch of minor stylistic cleaning and organizing, which is why it will show that a lot of files are changed, although most are just style. Also fixed a few minor bugs not worth describing in detail.

* `cmdstanr` now the default backend *if it is installed*. If it is not, `rstan` is used instead. 

Closes #142 
Closes #145 
Closes #72 (last comment)

Note that after this you will need to update the M3 code to ensure consistency with the new changes. Mainly:

- Specify the default_priors in the model object
- Do not construct priors in the configure_model.M3custom method
- Save the family into the formula and only return the formula but not the family from configure_model. I think this should be mostly it

#### Example

Here's an example of how the new features work and the output:

```r
formula <- bmf(mu ~ 1,                 # estimate mu, which is fixed to 0 by default
               c = 1.5,                # fix c to a constant and don't estimate it
               kappa ~ 0 + set_size)   # estimate kappa as usual
fit <- fit_model(formula, OberauerLin_2017, sdmSimple('dev_rad'), parallel = TRUE, backend = "cmdstanr")
```

image

#### Tests


venpopov:
I'll put comments in the code to orient to the main changes (I should have done the stylistic changes separately, now they make it difficult to see key changes unfortunately)

.github/workflows/R-CMD-check.yaml
venpopov, line 1: to make cmdstanr work on the automatic checks

R/bmm_model_IMM.R
```diff
@@ -36,30 +32,32 @@
...
-                     a = "identity",
-                     c = "identity"),
-        fixed_parameters = list(mu1 = 0),
+        links = list(
+          kappa = "log",
+          a = "identity",
+          c = "identity"
+        ),
+        fixed_parameters = list(mu1 = 0, mu2 = 0, kappa2 = -100),
+        default_priors = list(
```
venpopov, line 41: the default priors list is now part of the model object. Same for all models

R/bmm_model_IMM.R
```diff
@@ -36,30 +32,32 @@
...
-        links = list(kappa = "log",
-                     a = "identity",
-                     c = "identity"),
-        fixed_parameters = list(mu1 = 0),
+        links = list(
+          kappa = "log",
+          a = "identity",
+          c = "identity"
+        ),
+        fixed_parameters = list(mu1 = 0, mu2 = 0, kappa2 = -100),
```
venpopov, line 40: all fixed parameters, including the internal now are part of the model object. Necessary for the new prior method to work. They are still not printed in the summary

R/bmm_model_IMM.R
```diff
@@ -307,189 +309,162 @@ check_data.IMMspatial <- function(model, data, formula) {
...
-  kappa_nts <- paste0('kappa', 2:max_setsize)
-  kappa_unif <- paste0('kappa',max_setsize + 1)
-  theta_nts <- paste0('theta',2:max_setsize)
-  mu_nts <- paste0('mu', 2:max_setsize)
-  mu_unif <- paste0('mu', max_setsize + 1)
-
-  formula <- formula +
-    glue_lf(kappa_unif,' ~ 1') +
-    glue_lf(mu_unif, ' ~ 1') +
+  formula <- bmf2bf(model, formula) +
```
venpopov, line 317: simplified the formula construction for all non-target models. now the uniform distribution is kappa2 and mu2, which makes it much easier because their index does not depend on the setsize. Allows to build the constant priors with a general function outside for all models

R/bmm_model_IMM.R
```diff
@@ -307,189 +309,162 @@ check_data.IMMspatial <- function(model, data, formula) {
...
+  mu_nts <- paste0("mu", 3:(max_setsize + 1))
+
   for (i in 1:(max_setsize - 1)) {
     formula <- formula +
-      glue_nlf(kappa_nts[i], ' ~ kappa') +
-      glue_nlf(theta_nts[i], ' ~ ', lure_idx_vars[i], '*(a) + ',
-               '(1-', lure_idx_vars[i], ')*(-100)') +
-      glue_nlf(mu_nts[i], ' ~ ', nt_features[i])
+      glue_nlf("{kappa_nts[i]} ~ kappa") +
+      glue_nlf("{theta_nts[i]} ~ {lure_idx[i]} * a + (1 - {lure_idx[i]}) * (-100)") +
```
venpopov, line 331: I rewrote the glue_nlf function to allow us to use cleaner syntax - variables are specified between {} brackets, the rest is literal text

R/bmm_model_IMM.R
```diff
@@ -307,189 +309,162 @@ check_data.IMMspatial <- function(model, data, formula) {
...
-  additional_constants[[mu_unif]] <- 0
-  prior <- fixed_pars_priors(model, additional_constants)
-  if (getOption("bmm.default_priors", TRUE)) {
-    prior <- prior +
-      set_default_prior(bmm_formula, data,
-                        prior_list=list(kappa=list(main = 'normal(2,1)', effects = 'normal(0,1)', nlpar=T),
-                                        a=list(main = 'normal(0,1)', effects = 'normal(0,1)', nlpar=T),
-                                        c=list(main = 'normal(0,1)', effects = 'normal(0,1)', nlpar=T)))
-  }
+  formula$family <- brms::do_call(brms::mixture, vm_list)
```
venpopov, line 338: family now must be attached to the formula object

R/bmm_model_IMM.R
```diff
@@ -307,189 +309,162 @@ check_data.IMMspatial <- function(model, data, formula) {
...
+
+  nlist(formula, data)
```
venpopov, line 340: do not return family or prior any more

R/bmm_model_IMM.R
```diff
@@ -307,189 +309,162 @@ check_data.IMMspatial <- function(model, data, formula) {
...
+}
 
-  # if there is setsize 1 in the data, set constant prior over a for setsize1
-  a_preds <- rhs_vars(bmm_formula$a)
+
+#' @export
+configure_prior.IMMabc <- function(model, data, formula, user_prior, ...) {
```
venpopov, line 345: the only prior that needs to be defined specifically for each model is that which depends on model and data conditions, such as the setsize here. Everything else is handled by configure_prior.bmmmodel()

R/bmmformula.R
```diff
@@ -75,33 +74,41 @@
...
+      stop2("The arguments must be formulas or numeric values.")
+    }
+    stopif(par %in% names(formula), "Duplicated formula for parameter {par}")
+    formula[[par]] <- arg
   }
+  class(formula) <- "bmmformula"
   # assign attribute nl TRUE/FALSE to each component of the formula
-  assign_nl(formula)
+  formula <- assign_nl(formula)
+  assign_constants(formula)
```
venpopov, line 95: determines which parameters should be treated as constants rather than estimated

R/bmmformula.R
```diff
@@ -75,33 +74,41 @@
...
-  names(formula) <- all.vars(formula[[1]])[1]
-  class(formula) <- "bmmformula"
-  for (f in dots) {
-    formula <- formula + f
+  formula <- list()
+  for (i in seq_along(dots)) {
+    arg <- dots[[i]]
+    if (is_formula(arg)) {
+      par <- all.vars(arg)[1]
+    } else if (is.numeric(arg) && length(arg) == 1) {
```
venpopov, line 84: allows specifying "par = value", which is not actually a formula

R/distributions.R
venpopov, line 1: only style changes, no changes to functionality

R/fit_model.R
```diff
@@ -80,38 +84,32 @@ fit_model <- function(formula, data, model,
...
   formula <- check_formula(model, data, formula)
 
-
   # generate the model specification to pass to brms later
   config_args <- configure_model(model, data, formula)
 
-  # combine the default prior plus user given prior
-  config_args$prior <- combine_prior(config_args$prior, prior)
+  # configure the default prior and combine with user-specified prior
+  prior <- configure_prior(model, data, config_args$formula, prior)
```
venpopov, line 106: configure_prior is now the main way we construct:

- the default prior
- the constant priors
- combine with user prior

R/helpers-data.R
venpopov, line 1: only stule changes, can ignore

R/helpers-model.R
```diff
@@ -110,34 +89,35 @@ configure_model <- function(model, data, formula) {
...
+
+          {supported_models()}")
   model
 }
 
 #' @export
-check_model.bmmmodel <- function(model, data = NULL) {
+check_model.bmmmodel <- function(model, data = NULL, formula = NULL) {
   model <- replace_regex_variables(model, data)
+  model <- change_constants(model, formula)
```
venpopov, line 120: this line is key to making everything work - it will read the "constant" attribute from the formula and replace the fixed_parameters field in the model object. This information is then used by all other methods such as the summary, configure_prior, etc

R/helpers-parameters.R
venpopov, line 1: only style changes, can ignore

R/helpers-prior.R
```diff
@@ -90,151 +91,182 @@ get_model_prior <- function(object, data, model, formula = object, ...) {
 #'   class="Intercept", dpar=parameter_name) for all fixed parameters in the
 #'   model
 #' @noRd
-fixed_pars_priors <- function(model, additional_pars = list()) {
+fixed_pars_priors <- function(model, formula, additional_pars = list()) {
```
venpopov, line 94: this function now automatically constructs the constant priors, recognizing which parameters should nlpars, which column to put the intercept in, etc. It now takes the brmsformula that is returned from configure_model(), because we can easily extract that information from there

R/helpers-prior.R
```diff
@@ -90,151 +91,182 @@ get_model_prior <- function(object, data, model, formula = object, ...) {
...
-    stop("You are trying to set a default prior on a parameter that is not part of the model")
-  }
-  if (!is.list(prior_list)) {
-    stop("The prior_list should be a list of lists")
-  }
-  for (i in 1:length(prior_list)) {
-    if(!is.list(prior_list[[i]])) {
-      stop("The prior_list should be a list of lists")
-    }
+set_default_prior <- function(model, data, formula) {
```
venpopov, line 130: now takes the final brmsformula instead of the bmm formula, making it easier to determine which pars are nlpars for the priors without having to specify that manualyl in the priors_list as before

R/helpers-prior.R
```diff
@@ -90,151 +91,182 @@ get_model_prior <- function(object, data, model, formula = object, ...) {
...
+  UseMethod("configure_prior")
+}
+
+#' @export
+configure_prior.default <- function(model, data, formula, user_prior, ...) {
+  NULL
+}
+
+#' @export
+configure_prior.bmmmodel <- function(model, data, formula, user_prior, ...) {
```
venpopov, line 253: the main new S3 method - called by fit_model() after configure_model(). It flexibly constructs all the priors with very little code

R/utils.R
venpopov, line 1: only style changes, can ignore

R/bmm_model_IMM.R
```diff
@@ -36,30 +32,32 @@
...
-        fixed_parameters = list(mu1 = 0),
+        links = list(
+          kappa = "log",
+          a = "identity",
+          c = "identity"
+        ),
+        fixed_parameters = list(mu1 = 0, mu2 = 0, kappa2 = -100),
+        default_priors = list(
+          kappa = list(main = "normal(2,1)", effects = "normal(0,1)"),
+          a = list(main = "normal(0,1)", effects = "normal(0,1)"),
```
venpopov, line 43: the default priors list is now part of the model object. Same for all models

GidonFrischkorn (approved):
This looks all good to me. I think I have also grasped all the relevant changes and will apply them to the M3 models. Should I have any questions, I will get back to you. Thanks also for taking care of the cosmetic changes to the code.

venpopov:
Cool, thanks for reviewing it!

PR #162, opened 2024-03-22
Documentation/153 update oberauerlin 2017

GidonFrischkorn:
#### Summary
- Remove unnecessary values in the `OberauerLin_2017 `data set.
- Adapt `check_data.IMMspatial` to deal with `NA` values

#### Tests


R/bmm_model_IMM.R
```diff
@@ -293,6 +293,9 @@ check_data.IMMspatial <- function(model, data, formula) {
          "The number of columns for non-target distances in the argument \\
          'nt_distances' should equal max(setsize)-1})")
 
+  # replace nt_distances
+  data[,nt_distances][is.na(data[,nt_distances])] <- max(data[,nt_distances], na.rm = T)
```
GidonFrischkorn, line 297: This line deals with `NA` values in the `nt_distances` variables. Otherwise the code does not compile. We do the same for the `nt_features` in the `check_data.nontargets` function.

venpopov:
Cool thanks! Will review in the afternoon

R/bmm_model_IMM.R
venpopov: I'd suggest fixing it to a constant value, maybe a really high constant like 999, indicating these are special values. The actual value doesn't matter since lure_idx is 0 and we multiply by it, but it is just to reduce confusion - I was at first wondering why the max value is used. Otherwise looks good

venpopov (changes requested):

venpopov (approved):

PR #166, opened 2024-03-22
add print method for bmmformula, simplify summarise_formula

### LINKED ISSUE #165: Add `print.bmmformula` method to make it look nicer when the `bmmformula` is printed to the console
Similar to the print method used for `brmsformula` objects we should add a print method for `bmmformula` object.

venpopov:
#### Summary

Fixes #165 

#### Tests


GidonFrischkorn (approved):
Thanks for taking care of this so quickly!

PR #172, opened 2024-03-23
Ensure fit_model and default_prior work when there are no fixed parameters

### LINKED ISSUE #170: setting prior fails if there are no fixed parameters in the model
After #152, a new bug appears if we try to predict all parameters of a model and none are left fixed:

```r
formula <- bmf(mu ~ 1,
               c ~ 1,                
               kappa ~ 1)   
default_prior(formula, OberauerLin_2017, sdmSimple('dev_rad'))
```
```
Error: Processing arguments of 'set_prior' has failed:
Error in (function (..., row.names = NULL, check.rows = FALSE, check.names = TRUE,  : 
  arguments imply differing number of rows: 0, 1
```

If at least one parameter is fixed to a constant it works fine:

```r
formula <- bmf(mu ~ 1,
               c ~ 1,                
               kappa = 2)   
default_prior(formula, OberauerLin_2017, sdmSimple('dev_rad'))
```

This is caused by the "fixed_pars_priors", which throws an error if there are no fixed parameters. We should add a condition to execute this only if fixed parameters exist

venpopov:
#### Summary

Fixes #170 

#### Tests

### DIFF
--- a/R/bmm_model_IMM.R
+++ b/R/bmm_model_IMM.R
@@ -39,6 +39,7 @@
         ),
         fixed_parameters = list(mu1 = 0, mu2 = 0, kappa2 = -100),
         default_priors = list(
+          mu1 = list(main = "student_t(1, 0, 1)"),
           kappa = list(main = "normal(2,1)", effects = "normal(0,1)"),
           a = list(main = "normal(0,1)", effects = "normal(0,1)"),
           c = list(main = "normal(0,1)", effects = "normal(0,1)")
@@ -90,6 +91,7 @@
                      s = "log"),
         fixed_parameters = list(mu1 = 0, mu2 = 0, kappa2 = -100),
         default_priors = list(
+          mu1 = list(main = "student_t(1, 0, 1)"),
           kappa = list(main = "normal(2,1)", effects = "normal(0,1)"),
           c = list(main = "normal(0,1)", effects = "normal(0,1)"),
           s = list(main = "normal(0,1)", effects = "normal(0,1)")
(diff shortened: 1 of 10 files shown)


R/bmm_model_IMM.R
```diff
@@ -39,6 +39,7 @@
         ),
         fixed_parameters = list(mu1 = 0, mu2 = 0, kappa2 = -100),
         default_priors = list(
+          mu1 = list(main = "student_t(1, 0, 1)"),
```
GidonFrischkorn, line 42: If this can be estimated freely, maybe some users might want to estimate if biases differ between conditions. In these cases it would make sense to also provide an effect prior. I assume these will be very few cases, but I still think it is better to specify one, should a user use this feature.

GidonFrischkorn (changes requested):
I just had one comment, to add an effect prior for the `mu1` parameters. Technically, users can specify effects on this parameter. So, we should add a default prior for effects too.

R/bmm_model_IMM.R
GidonFrischkorn: I see. Then, let's do it for now without the effect prior. But I think, we should add this info in the vignettes or wherever we explain how internally fixed parameters can be estimated.

venpopov: Agreed, we should explain it 👍

GidonFrischkorn: Perfect. If you re-invite my review, then I will approve these changes and we take care that we communicate that currently no priors are specified on effects for `mu` parameters in the vwm models.

GidonFrischkorn (approved):

PR #174, opened 2024-03-23
Remove deprecated functions; hard dependency on brms 2.21.0

### LINKED ISSUE #168: Clean up code after the release of brms 2.21.0 is complete
Paul has initiated the release of 2.21.0 on CRAN, so in a couple of days when it is done, we can get rid of all the conditional statements and make this brms version a hard dependency.

Tasks:
- [x] rename all uses to stancode, standata and default_prior
- [x] remove conditionals in the restructure method and summary
- [ ] write a vignette about these functions #146

venpopov:
#### Summary

Closes #168 

Now brms 2.21.0 is a hard dependency, and we get rid of all the older functions. Only default_prior(), stancode(), standata() remain.

#### Tests


GidonFrischkorn (approved):
Looks good to me. I checked it on my machine and adapted some of the test code I had and it ran all fine. Thanks for cleaning this up!

PR #175, opened 2024-03-23
Documentation/100 vignette bmmformula

GidonFrischkorn:
#### Summary
- Add vignette for `bmmformula`

#### Tests


GidonFrischkorn:
This is a first but pretty comprehensive draft. I wanted to share this and ask you for comments. So, feel free to comment and make edits as you see fit.

venpopov:
> This is a first but pretty comprehensive draft. I wanted to share this and ask you for comments. So, feel free to comment and make edits as you see fit.

It looks great! Thanks for putting this together.

The one suggestion I have is to skip the section "Using constant priors". I think it only adds unnecessary complexity without offering any additional features.

In it the call to default prior should be default_prior(my_formula, my_data, my_model), but if you remove the section that doesn't matter.

vignettes/bmmformula.Rmd
```diff
@@ -0,0 +1,185 @@
...
+
+To fix a parameter via the `bmmformula` you have to set the parameter to the value it should be fixed to by using `=` instead of `~`. For example, the following formula would fix `kappa` to the value `1`.
+
+```r
+fixed_formula <- bmf(kappa = 1)
+```
+
+When fixing a parameter to a constant value, you need to keep in mind that the value you fix it to, is the value on the parameter scale. Oftentime, `bmm` will use link functions to convert parameters from a bounded native scale (e.g. only positive values for precision parameters, such as `kappa`) to an unbounded parameter scale. In the case of `kappa`, `bmm` uses a `log` link function. Thus, when fixing the parameter to `1`, this will effectively fix the parameter to $e^1$ on the native scale.
+
+## Using `constant` priors
```
venpopov, line 120: as I noted in the PR reply, I think this should be removed

vignettes/bmmformula.Rmd
```diff
@@ -0,0 +1,185 @@
...
+                     set_size = "ss")
+```
+
+Internally, `bmm` will then link the response variables to the `bmmformula` given for the parameters of a `bmmmodel`.
+
+# Fixing parameters to constant values
+
+In some cases it can be reasonable to fix parameters to a specific value. This can either be done via the `bmmformula` or via a constant prior specified for the parameter to be fixed. 
+
+## Inside the `bmmformula`
```
venpopov, line 110: if you remove the constant priors section, just remove this subheading

vignettes/bmmformula.Rmd
```diff
@@ -0,0 +1,185 @@
...
+  # parameters of the non-linear function
+  start_theta ~ 1 + (1 | gr(ID, by = age_group)),
+  end_theta ~ 1 + (1 | gr(ID, by = age_group)),
+  R ~ 1 + (1 | gr(ID, by = age_group))
+)
+```
+
+This formula implements an exponential reduction in `thetat` from a `start_theta` at `ss_num = 0` towards the lower asymptote `end_theta` that will be reached with the slope `R`. In addition, the group level effects have been group by `age_group` so that the variation in the `Intercept` would be estimated seperately for different levels of the `age_group` variable. This formula just serves as an example how `bmmformula` syntax allows to use the complex formula syntax implemented in `brms` and passes the information on to `brms` without losing any of the functionality. 
+
+## Seperating response variables from model parameters
```
venpopov, line 94: make heading 1st level

vignettes/bmmformula.Rmd
```diff
@@ -0,0 +1,185 @@
...
+constant_prior <- prior(constant(1), coef = "Intercept", dpar = "kappa")
+
+# for nlpars
+constant_prior <- prior(constant(1), coef = "Intercept", nlpar = "kappa")
+```
+
+# Accessing the `brmsformula` generated by `bmm`
+
+```{r, include=FALSE, warning=FALSE, message=FALSE}
+bmmfit <- bmm:::bmmfit_mixture2p_vignette
```
venpopov, line 147: if you keep this hidden, I suggest you include the code to fit the model from the other vignette. Right now someone might wonder where this bmmfit came from and be confused about where to extract the info from

vignettes/bmmformula.Rmd
```diff
@@ -0,0 +1,185 @@
...
+
+Finally, users can also access the data that was used by `bmm` to fit the model via `bmmfit$data`. In many cases this data will be equal to the data provided by the user. But sometimes `bmm` internally computes additional index variables for specifying the models adequately. The data stored in the `bmmfit` object contains these additional variables.
+
+```{r}
+head(bmmfit$data)
+```
+
+This way users who are interested to customize models to fit them in `brms` without `bmm` are able to obtain the most important information that is essential for specifying the models implemented in `bmm`. 
+
+Likewise users can call a `brms` summary instead of a `bmm` summary by setting the `backend` option in the summary function to `brms`. This way the summary function for `brmsfit` objects will be used instead of the function for `bmmfit` objects. The `brmsfit` summary also contains the `brmsformula` that is created by `bmm`.
```
venpopov, line 181: Add sentence for context: "bmm model use a custom summary function to format the output. You can call a ...."

venpopov (changes requested):
I left a few specific comments, but overall it lookgs great. An overall comment is inconsistent use of second and third person language - "you can" vs "users can" vs "we can access". I suggest sticking with either you or we, and avoid the third person.

venpopov:
Another useful thing to add would be that people can choose to estimate some of the parameters that are fixed by default

GidonFrischkorn:
I have included your comments and also added a section on estimating internally fixed parameters. I added some information there how to access model parameters and distinguish parameters fixed for identification purposes from fixed parmaeters that can be estimated if needed.

Feel free to add further comments, and I hope that I managed to consistently use "you" in this version of the vignette.

venpopov (approved):

venpopov:
A few minor edits from my side, and fixed some code I noticed from the summaries in the vignettes. Good to go

venpopov:
The failed checks seem like they have nothing to do with us. I looked into it, and apparently this can happen when a dependency is updated on CRAN and it takes a few days to stabilize... https://stackoverflow.com/questions/51453717/issue-with-r-package-check-on-windows-package-suggested-but-not-available-str

Issue #163, opened 2024-03-22
conventions on variables, arguments, functions, etc


bnicenboim:
Hi,
As I said before, amazing idea! I'm sorry to be annoying with this (which seems not that important, but in my experience it is). But it would be good to have some wiki/contribution guidelines for how to name arguments, functions, and variables consistently. This will help (me/others) to contribute new models and to start using the existing functions without the worry that they will change (a lot, of course they'll change it's in development):

I've noticed that there are already some inconsistencies, so for example here:

`IMMfull(resp_err = "dev_rad",
                 nt_features = paste0("col_nt", 1:7),
                 nt_distances = paste0("dist_nt",1:7),
                 setsize = "set_size")`

Shouldn't it be `set_size`? Does it make sense to cut `resp_error` into `resp_err`? Also, how are models named? Given that you have `IMMfull` (why not `IMM_full`?), shouldn't it be `SDMsimple` (or better `SDM_simple`) rather than sdmSimple?

bnicenboim:
This might be too related to #134. But my main point would be to have contribution guidelines.

GidonFrischkorn:
Hi Bruno,
thanks for bringing this point up, and yes this is something we should further clarify in the developer notes! There have been some changes that are not yet reflected there, so they need to be updated ASAP anyway. Generally, I agree that we should look over the naming scheme and use snake_case consistently. 

So for example `setsize` should be `set_size`, comparable to `nt_features` and I also see that we could change `resp_err` to `resp_error` as it does not really lead to a significantly longer variable name.

Regarding the model naming, there is a clear tendency for @venpopov and me to go for an additional `version` argument inside the model call. From our perspective this would clarify that the same model can be applied in different tasks or flavors depending on the assumptions a researcher wants to make.

Should you have more feedback, feel free to bring it up here. We are happy about everything that helps getting this package to an easy to use and flexible tool that as many people as possible can contribute to.

venpopov:
I agree. This has bugged me for a while and a few weeks ago I changed all internals to underscore conventions, and was planning to raise an issue for the user facing functions as well.

My preference would be for everything to be using underscores and lower case. sdm_simple, etc. @GidonFrischkorn thoughts?

venpopov:
Posted this before seeing your comment above. Yes, I'm forr the version argument we discussed, but for everything else UAE that naming consistency

venpopov:
[...]

PR #177, opened 2024-03-25
Feature/issue 163 naming consistency

venpopov:
#### Summary

This PR changes the easier things from #163:

- fit_model() -> bmm()
   - deprecated but can still use fit_model() with a warning message
- dIMM and friends -> dimm
   - hard deprecation
- setsize argument -> set_size
  - deprecated but still usable with a warning
- OberauerLin_2017 -> oberauer_lin_2017
- ZhangLuck_2008 -> zhang_luck_2008
- resp_err -> resp_error
- deprecate parallel option
   - still usable with a warning, but should just use `cores = N` where N is the number of chains to run in parallel, just like in brms
- bmmmodel class -> bmmodel

Since restructuring the models to use a version argument would require a bigger change and testing, I decided to make this into a separate PR.

#### Tests


GidonFrischkorn (approved):
Phew... this was tracking a lot of find & replace details. I tested the changes and added warnings you implemented on my local testing codes and everything ran as expected. This has reduced inconsistencies throughout the package a lot, but it is likely we will still find some every now and then.

PR #178, opened 2024-03-26
Restructure models to use a "version" argument instead of separate functions

### LINKED ISSUE #163: conventions on variables, arguments, functions, etc

venpopov:
#### Summary

This is the follow up to #177. Closes #163 

the sdmSimple() model is now sdm()
the IMMfull(), IMMabc() and IMMbsc() are now all imm(), with a version argument:
- imm(version = "full")
- imm(version = "abc")
- imm(version = "bsc")

the "full" is the default

I decided to leave the two mixture models separately. For them almost all the meta information is different - the model name, the citation, etc. So I thought it's easier to keep the documentation separately, otherwise the help page becomes complicated.

This framework should now work for future models that have multiple versions.

#### Tests

I reran all the internal reference fits tests as this is a more substantial change. All is good and we replicate previous fits.


tests/testthat/test-fit_model.R
GidonFrischkorn, line 1: Should we rename this file to `test-bmm.R` ? I think this would just be for consistency in naming, but it might still be good to just take care of this now.

R/model_IMM.R
```diff
@@ -162,25 +55,43 @@
       # attributes
       regex = regex,
       regex_vars = c('nt_features', 'nt_distances'),
-      class = c("bmmodel", "vwm", "nontargets", "IMMspatial", "IMMfull")
+      class = c("bmmodel", "vwm", "non_targets", "imm", paste0('imm_',version)),
```
GidonFrischkorn, line 58: Here, we need another class that is shared between the `imm_full` and `imm_bsc` that we called `imm_spatial`. There were some shared data checks between these that we collected there.

R/model_IMM.R
```diff
@@ -306,7 +262,18 @@ IMMabc <- function(resp_error, nt_features, set_size, regex = FALSE, links = NUL
...
+  NextMethod("check_data")
+}
+
+#' @export
+check_data.imm_full <- function(model, data, formula) {
+  data <- .check_data_imm_dist(model, data, formula)
+  NextMethod("check_data")
+}
+
+.check_data_imm_dist <- function(model, data, formula) {
```
GidonFrischkorn, line 276: Typo in `.check_data_imm_dist`. This needs to be `.check_data.imm_dist` Accordingly, the additional class for the `imm_full` and the `imm_bsc` using the `nt_distances` should be declared as `imm_dist`.

GidonFrischkorn (changes requested):
I found some minor things in the `model_imm.R` script that I commented. Apart from that I did not run into any issues on my machine and like the cleaner setup for the different models. I also think this will make it easier to accomodate different task versions for these models (e.g. for change detection versus continous reproduction).

Thanks for taking care of this.

GidonFrischkorn:
In [b3e85cb](https://github.com/venpopov/bmm/pull/178/commits/b3e85cb39c84efaa4c47e2d211fc864c2426b5b1) I added the changes to the  `use_model_template`. You might want to check these changes and look if there is something missing.

R/model_IMM.R
venpopov: No, if you look above I just call this regular function from the check_data.imm_full and check_data. Imm_bsc, so no need for the extra class. I found this way a bit cleaner since we don't need now an if else statement in model_imm to add a special spatial class

R/model_IMM.R
venpopov: See my comment on the data check function, I just decided to do it in a way that we don't need the extra class

R/model_IMM.R
GidonFrischkorn: Ah, now I see what you did! Sorry for not getting it the first time...

GidonFrischkorn (approved):

R/model_IMM.R
venpopov: no problem, I should have noted that change in the PR message!

venpopov:
> In [b3e85cb](https://github.com/venpopov/bmm/pull/178/commits/b3e85cb39c84efaa4c47e2d211fc864c2426b5b1) I added the changes to the `use_model_template`. You might want to check these changes and look if there is something missing.

Looks good! Thanks!

PR #207, opened 2024-04-07
Bugfix for setsize1 ranef constant prior

### LINKED ISSUE #202: [bug] `sd` for `thetant` for setsize = 1 not fixed to zero in `mixture3p`
**Describe the bug**
When working on the revisions of the tutorial paper I noticed that when we estimate the `mixture3p` using random effects for the `set_size` variable that there are still random effects estimated for `thetant` for set size = 1. See screenshot below

image

**To Reproduce**
You can see the specified call to `bmm` in the screenshot of the summary. 

**Expected behavior**
Ideally, we should fix the `sd` to zero and list them under `Constant Parameters`. But I do not know if this is possible via priors. There might be constraint that `sd` parameters cannot be fixed to zero, but only a value very close to zero.

venpopov:
#### Summary

Fixes #202 by placing constant(1e-8) prior on sd for thetant, a or s for mixture2p or imm for coefficient set_size1


GidonFrischkorn (approved):
Looks good. With the examples I have I did not run into any problems. Thanks for fixing this, despite it being more cosmetic than substantial.

venpopov:
This one actually branched off from the other, so merging this should be sufficient

PR #212, opened 2024-05-21
update vignettes and delete internal datasets

venpopov:
#### Summary

Closes #185. We include vignettes only as articles on the website. The internal datasets for them are no longer included in the package to reduce package size for CRAN

#### Tests


venpopov:
I just realized that we have several places in the documentation that we write "check `vignette('....')`. Should replace these with links to the website instead now that the vignettes are not bundled with the package. So I'm marking this as a draft until I do that

GidonFrischkorn (approved):

GidonFrischkorn (approved):

PR #229, opened 2024-08-14
Feature/224 generalize wrap deg2rad

### LINKED ISSUE #224: [feature-request] generalize the `wrap` and `deg2rad` functions for half-circular feature spaces (e.g. bar orientations)
In some visual WM tasks participants are asked to remember the orientation of a bar. This feature space does only encompass a half-circle and thus we would have to adapt some of the helper function for the `circular` model. The user could just multiply the deviation in degrees with 2 and thereby rescale the feature space to a full-circle, but I wonder if we could handle that more elegantly with some of the helper function implemented in the package.

The other option would be to add some information in the article on continuous reproduction tasks that explains how to treat half-circular feature spaces when using the respective `bmmodels`

What do you think @venpopov?

GidonFrischkorn:
#### Summary
- This is a proposition for adding a scale argument to the `circle_transform` functions. This would generalize them towards using them with bar orientations that have a scale of 180 degrees and also any other circular space with a different circle circumference.

If you think it is not really necessary to generalize this, I do not think this is super important. But I had a look into this a while ago and now wanted to share it, so you can say what you think.

#### Tests

#### Release notes
- you can now pass the **scale** to the **circle_transform** functions, so that these functions also work with circular spaces beyond 360 degrees


R/helpers-data.R
```diff
@@ -213,37 +216,39 @@ calc_error_relative_to_nontargets <- function(data, response, nt_features) {
 #' wrapped_diff <- wrap(x - y)
 #' hist(wrapped_diff)
 #'
-wrap <- function(x, radians = TRUE) {
+wrap <- function(x, radians = TRUE, scale = 360) {
   stopifnot(is.logical(radians))
   if (radians) {
     return(((x + pi) %% (2 * pi)) - pi)
```
venpopov, line 229: this also needs to be corrected

venpopov:
Sorry, I somehow had missed the issue you previously created!
I'm conflicted about this. If we implement the scale argument to `wrap` and `deg2rad` that's fine, but to be useful we would need to add is an argument to all VWM-CRT models so that it will be applied during the model checks. 

Crucially, we would also need to add the scale argument to the data because the von_mises distribution will not model the data correctly because it expects a complete circular space. So we will have to multiply the angle anyway. And communicate to the user that we have done so, and that they need to take that into account when interpreting the precision estimates, posterior predictive checks, etc. I think this would be a mess.

At this point it would be just much easier to leave that to the user, who would just need to multiply their angles, as you say.

GidonFrischkorn:
No worries. It was just a small thing that popped into my head and then I quickly looked how to generalize it.

I see your point that this has the potential to confuse users and I agree that it is not ideal to completely integrate this as an additional argument in the `bmm` call for the circular-VWM models. The longer I think about this I see your point and maybe we should just leave the functions the way they are.

Maybe, we can add one or two sentences to the article on the continuous reproduction task, that orientations without a direction (bars, gabor patches) need to be multiplied by two, to accommodate for the half-circular space they are in.

venpopov:
> Maybe, we can add one or two sentences to the article on the continuous reproduction task, that orientations without a direction (bars, gabor patches) need to be multiplied by two, to accommodate for the half-circular space they are in.

This is a good idea!

venpopov:
I'm closing this so that it doesn't show up in the list of PRs and I opened an issue to remind us to add this info the article

PR #232, opened 2024-08-21
Priors nlformula

GidonFrischkorn:
#### Summary
- Remove default priors for model parameters that are transformed via non-linear functions
- Print warning if default priors are removed, to notify users that they should specify proper priors themselves

#### Tests

#### Release notes
- Fix a conflict in setting default priors when model parameters were transformed in a non-linear formula


GidonFrischkorn:
One little thing I do not like currently, is that the warning for the missing default priors is printed for all functions that call the `check_model.bmmodel` method. That is when you call `standata` or `stancode` you get the warning also you might not be doing something that is related to priors.

I wondered if there is some option to access by which function call the `check_model.bmmodel` method was called and only print the warning for calls by `default_prior` and `bmm` (maybe also `stancode` as the priors are also included there.) If you have an idea how to implement that I think adding this would be great.

venpopov:
Thanks for solving this. I made a couple of changes:
- moved the check for default priors into the `set_default_priors` function instead of the `check_model` function
- removed prior specification from the standata() function - that way with the first change you get no warning when using standata
- added a test for this issue to make sure it works and we don't break it in the future

venpopov (approved):

PR #234, opened 2024-08-23
align `file_refit` options with brms

GidonFrischkorn:
#### Summary
- Allow "never" and "always" to be passed to `file_refit`. Throw a warning if "on_change" is passed

#### Tests

#### Release notes
* the `file_refit` argument of the `bmm` function now accepts character string like brms. A warning is given when "on_change" is specified, as this is not currently implemented for `bmmodels`.

### DIFF
--- a/R/bmm.R
+++ b/R/bmm.R
@@ -45,11 +45,11 @@
 #' @param file_compress Logical or a character string, specifying one of the
 #'   compression algorithms supported by [saveRDS] when saving
 #'   the fitted model object.
-#' @param file_refit Logical. Modifies when the fit stored via the `file` argument is
+#' @param file_refit Logical or character string. Modifies when the fit stored via the `file` argument is
 #'   re-used. Can be set globally for the current \R session via the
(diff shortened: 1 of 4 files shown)


R/bmm.R
```diff
@@ -45,11 +45,11 @@
 #' @param file_compress Logical or a character string, specifying one of the
 #'   compression algorithms supported by [saveRDS] when saving
 #'   the fitted model object.
-#' @param file_refit Logical. Modifies when the fit stored via the `file` argument is
+#' @param file_refit Logical or character string. Modifies when the fit stored via the `file` argument is
```
GidonFrischkorn, line 48: Can you check if these changes are correct! I was a little confused because previously the documentation states that `file_refit = TRUE` would lead to re-using a saved model, and `file_refit = FALSE` would refit the model. I would have expected to be exactly the other way around, and I think this is also what is implemented in the `read_bmmfit` function.

venpopov (approved):
Looks good. The previous documentation was incorrect

PR #242, opened 2025-01-22
Spring cleaning and refactoring

### LINKED ISSUE #238: Clean up and refactoring
As I prepare for reviewing #237 I notice minor things in the codebase that can be improved. It’s been a few months since we worked with the package so bringing some fresh eyes helps me notice some things. I’ll use this issue to make notes.

- [x] several occurances of `message2` use outdated paste syntax and should be replaced
- [x] ensure consistent spacing style - some old leftovers like in https://github.com/venpopov/bmm/blob/7c8c33c8e24f2a3f8e2c2d469d2c2140b78861a9/R/distributions.R#L545

### LINKED ISSUE #239: Remove unnecessary checks
I don’t think this check is necesary as the function is a bmmformula method which guarantees the first argument would be a bmmformula.

https://github.com/venpopov/bmm/blob/7c8c33c8e24f2a3f8e2c2d469d2c2140b78861a9/R/bmmformula.R#L108

- [x] double check if I had some reason to include that check that I don’t remember
- [x] check for other unnecessary checks

### LINKED ISSUE #240: Check for ununsed, rarely used and duplicate functions
I want to simplify the codebase and make it easier to work with.

- [x] Duplicate definitions of has_intercept

https://github.com/venpopov/bmm/blob/7c8c33c8e24f2a3f8e2c2d469d2c2140b78861a9/R/bmmformula.R#L294

https://github.com/venpopov/bmm/blob/7c8c33c8e24f2a3f8e2c2d469d2c2140b78861a9/R/brms-misc.R#L151

how has this not been flagged by build checks? Need to carefuly tests what happens

venpopov:
#### Summary

I hadn't looked at the code in months. While I was reviewing #237, I saw a lot of things in the existing code that I wanted to refactor with some things I've learned over the last half-year. I haven't changed any functionality, just streamlined the code and broke up some functions that were too big and convoluted (mostly breaking up deeply nested if/else and for loops). I also got rid of the dplyr and tidyr dependencies, which we used in only a couple of places. I've added a bunch of extra tests and removed comments which were outdated and did not reflect what the code was actually doing. Changed a bunch of variable names so that the intent is clear without the need for comments.

All the tests pass and there is no substantial new functionality. If you would like to run a couple of models you already have to make sure nothing breaks on your end. 

I will continue now the review of #237, but the changes here would need to be merged there as well.

Closes #238, closes #239, closes #240

#### Tests


venpopov:
@GidonFrischkorn I need to resolve some conflicts due to merging the M3 into develop, but after that it should be good to look at.

venpopov:
> @GidonFrischkorn I need to resolve some conflicts due to merging the M3 into develop, but after that it should be good to look at.

@GidonFrischkorn this is good for review now

R/bmmformula.R
```diff
@@ -73,25 +73,21 @@
...
-    }
-    stopif(par %in% names(formula), "Duplicated formula for parameter {par}")
-    formula[[par]] <- arg
-  }
-  class(formula) <- "bmmformula"
-  # assign attribute nl TRUE/FALSE to each component of the formula
-  formula <- assign_nl(formula)
-  assign_constants(formula)
+  out <- list(...)
+  is_form_or_num <- function(x) is_formula(x) || (is.numeric(x) && length(x) == 1)
```
GidonFrischkorn, line 77: Is there a specific reason for defining this function within the `bmmformula` function? Would it not be better to define it as an internal function outside of the `bmmformula`.

I assume there is a reason in that this formula is only used inside this formula.

GidonFrischkorn (approved):
I had one question regarding a function defined with the `bmmfunction`, but apart from that everything looks fine and I had no issues running some of my test scripts.

R/bmmformula.R
venpopov: That's exactly the reason. It is a strange condition that is unlikely to occur in many places. In principle we don't need a named function at all - we could just use an anonymous function directly in the sapply call right below it. This is just a new trick I learned, where instead of writing comments that explain what the code does, which my become outdated, is to write the code in a way that says what it does directly. In this particular case instead of reading this:

```r
  stopif(
   !all(sapply(out, function(x) is_formula(x) || (is.numeric(x) && length(x) == 1))), 
   "Arguments must be formulas or numeric assignments."
  )
  ```
  
and then trying to deparse what the logic is, the function name tells you what condition it checks for.

In this case it's not a huge difference, but it's a nice concept I learned about and I'm trying out in the code in a few places

Issue #244, opened 2025-01-28
[bug] Customize M3 using nonlinear regression formulas


chenyu-psy:
I am attempting to replicate the results of Exp 4 from the study by Oberauer and Lewandowsky (2019) using the same M3 in the `bmm` package. I first attempted to implement this non-linear regression model, but it returned the message: "You have not provided link functions for any of the specified parameters. Please provide link functions for all model parameters to ensure proper identification of your model." In the model, `i` and `t` represent data variables.
```r
Correct ~ e * a + e * c + b
Other ~ e * a + b
Dist ~ pow(f, i) * r * e * a + pow(f, i) * r * e * c + b
OtherDist ~ pow(f, i) * r * e * a + b
NPL ~ b
e ~ 1 + be * t
r ~ exp(-br * (t+0.5))
a ~ 1 + (1 || participant)
c ~ 1 + (1 || participant)
br ~ 1 + (1 || participant)
be ~ 1 + (1 || participant)
f ~ 1 + (1 || participant)
``` 

To address this issue, I adjusted the model accordingly; however, I still received another error message: 
"Error in str2lang(x) : <text>:2:0: unexpected end of input
1: Dist~pow( inv_logit(f) ,i)* inv_logit(r) *(1+ exp(e) *t)*a+pow( inv_logit(f) ,i)* inv_logit(r) *(1+
   ^"
```r
Correct ~ (1 + e * t) * a + (1 + e * t) * c + b
Other ~ (1 + e * t) * a + b
Dist ~ pow(f, i) * r * (1 + e * t) * a + pow(f, i) * r * (1 + e * t) * c + b
OtherDist ~ pow(f, i) * r * (1 + e * t) * a + b
NPL ~ b
a ~ 1 + (1 || participant)
c ~ 1 + (1 || participant)
r ~ 1 + (1 || participant)
e ~ 1 + (1 || participant)
f ~ 1 + (1 || participant)
``` 

Do you have any ideas for these issues?

venpopov:
can you post a short reproduicble code so that I can look into it?

chenyu-psy:
Yes, for sure! Please find the code below, and the data is attached.

```r

# package list
packages = c("brms", "bmm", "tidyverse")
# get the package names that are not installed
missing_packages <- setdiff(packages, rownames(installed.packages()))
# install the missing packages
if (length(missing_packages) > 0) install.packages(missing_packages)
# library package
suppressMessages(sapply(packages, library, character.only = TRUE))

[...]

PR #245, opened 2025-01-28
Add check for variables in data in `check_model.m3_custom`

GidonFrischkorn:
#### Summary
- Resolve issue with including variables contained in the data in non-linear formula for the `m3_custom` version of the `m3`
- The `check_model.m3_custom` method now also tests if any variables used in non-linear formulas are contained in the data and will not include those into the model parameters

#### Tests

#### Release notes
- I think we do not need to add something to the release notes for this fix as the `m3` is not yet included in the package.

### DIFF
--- a/R/bmmformula.R
+++ b/R/bmmformula.R
@@ -458,7 +458,7 @@ apply_links <- function(formula, links) {
   if (length(links) == 0) {
     return(formula)
   }
-  
+
   pars_with_links <- names(links)
   replacements <- sapply(pars_with_links, function(par) {
     switch(links[[par]],
@@ -469,12 +469,12 @@ apply_links <- function(formula, links) {
       stop2("Unknown link type {links[[par]]}. Check ?apply_links for supported types")
     )
   })
-  
+
   tbr_patterns <- paste0("\\b", pars_with_links, "\\b") # match whole word only
   names(tbr_patterns) <- pars_with_links
 
   for (dpar in names(formula)[is_nl(formula)]) {
-    formula_str <- deparse(formula[[dpar]])
+    formula_str <- deparse(formula[[dpar]], width.cutoff = 500L)
     for (par in pars_with_links) {
       formula_str <- gsub(tbr_patterns[par], replacements[par], formula_str)
     }
--- a/R/model_m3.R
+++ b/R/model_m3.R
@@ -121,7 +121,7 @@
 #'  `ss`, `cs`, or `custom`. The default is `custom`.
 #' @param ... used internally for testing, ignore it
 #' @return An object of class `bmmodel`
-#' 
+#'
 #' @details `r model_info(.model_m3(), components =c('domain', 'task', 'name', 'citation'))`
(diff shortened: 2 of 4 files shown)


GidonFrischkorn:
There is still one remaining issue.

When using the alternative specification that @chenyu-psy shared 

```
formula_M3_Exp4_alternative <- bmf(
  Correct ~ (1 + e * t) * a + (1 + e * t) * c + b,
  Other ~ (1 + e * t) * a + b,
  Dist ~ pow(f, i) * r * (1 + e * t) * a + pow(f, i) * r * (1 + e * t) * c + b,
  OtherDist ~ pow(f, i) * r * (1 + e * t) * a + b,
  NPL ~ b,
  a ~ 1 + (1 || participant),
  c ~ 1 + (1 || participant),
  r ~ 1 + (1 || participant),
  e ~ 1 + (1 || participant),
  f ~ 1 + (1 || participant)
)

model_M3_Exp4_alternative <- m3(
  resp_cats = c("Correct","Other","Dist", "OtherDist","NPL"),
  num_options = c("nCorrect","nOther","nDist","nOtherDist","nNPL"),
  choice_rule = "softmax",
  links = list(
    a = "identity",
    c = "identity",
    f = "logit",
    r = "log",
    e = "log"),
  default_priors = list(
    a = list(main = "normal(3,1)", effect = "normal(3,1)"),
    c = list(main = "normal(3,1)", effect = "normal(3,1)"),
    f = list(main = "normal(0,1)", effect = "normal(0,1)"),
    r = list(main = "normal(0,0.5)", effect = "normal(0,0.5)"),
    e = list(main = "normal(0,0.5)", effect = "normal(0,0.5)")
  )
)

default_prior(formula_M3_Exp4_alternative, data = mData_Exp4, model = model_M3_Exp4_alternative)
```

The `check_formula.bmmformula` method throws a warning when it calls `NextMethod(check_formula)` after the `check_formula.bmmodel` method and passing on to `check_formula.m3_custom`.

```
Warning message:
In formula.character(object, env = baseenv()) :
  Using formula(x) is deprecated when x is a character vector of length > 1.
  Consider formula(paste(x, collapse = " ")) instead.
```

This does not break the function, but I still do not understand what the problem is. Fitting the model is also no problem, but maybe you @venpopov have a better idea why this warning could pop up, when `NextMethod(check_formula)` in this transition.

venpopov:
omg, that was so stupid. `deparse`, which we use to convert a formula to a string, by default cuts strings longer than 60 characters into multiple strings. I never knew this, and it's such a strange default behavior. This resulted in the dist formula being split into multiple strings which in turn cause the warnings. I change the default to "500 characters", which is the max.

venpopov:
which now makes @chenyu-psy's error reported in #244 make sense - the dist formula was being cut which cause the unexpected end of input. But why that caused an error for him and only a warning for you, I can only guess....

venpopov (approved):
just add a test case for the bug, otherwise all good

PR #256, opened 2025-02-05
Add M3 example & adapt priors for custom version

GidonFrischkorn:
#### Summary
- Add example for `m3`
- Update default prior settings for `custom` version of the `m3`
- minor cleaning in the `bmf2bf.m3` method to allow for easier maintenance

#### Tests


GidonFrischkorn:
One minor thing that is still on my mind is that when the `simple` choice rule is used with `identity` link functions for the activations, the call to `brm` should include  `init = 0` to start sampling without warnings. Otherwise, there will be a lot of warning messages that negative probabilities were passed to the multinomial likelihood function.

I think we should print a warning if this is triggered, and I was not sure if I can just add this in the `configure_model.m3` method and have this passed onto the `config_args` in the `bmm` function.

venpopov:
Nice. I pushed some very minor clean-ups, nothing substantial. For the init, yes you can pass this. See how I did it for the sdm. You'll just have to make it conditional on the situation you described.

venpopov (approved):

R/model_m3.R
```diff
@@ -185,17 +223,25 @@ check_model.m3_custom <- function(model, data = NULL, formula = NULL) {
...
-    switch(model$links[[m]],
-      log = list(main = "normal(1,1)", effect = "normal(0,0.5)"),
-      identity = list(main = "normal(0,1)", effect = "normal(0,1)"),
-      logit = list(main = "normal(0,1)", effect = "normal(0,1)"),
-      stop2("Invalid link function provided! Please use one of the following link functions: identity, log, logit")
-    )
+    if (model$other_vars$choice_rule == "simple") {
+      switch(model$links[[m]],
+             log = list(main = "normal(1, 1)", effects = "normal(0, 0.5)"),
+             identity = list(main = "normal(10, 4)", effects = "normal(0, 1)"),
```
venpopov, line 233: `normal(10, 4)` - is this intentional or a typo?

PR #259, opened 2025-02-08
Abstract random generation in a rejection_sampling function

venpopov:
#### Summary

All the various random generation functions for our special distributions used the same logic - rejection sampling. Here I just abstracted that functionality into a new exported function "rejection_sampling" which now each function like rsdm, rmixture2p, rmixture3p and rimm uses. No changes to the interface (except for having a new function) - the remaining r* functions are used as before

#### Tests


GidonFrischkorn (approved):
Found no issues when running some small test. So good from my perspective.

PR #260, opened 2025-02-10
Fix problem with integer values passed to num_options

GidonFrischkorn:
- fix problem with creating index variables when integer values are passed
- add test that passing 0 in the integer vector produces an error

#### Tests


venpopov (approved):
I simplified some things. The previous logic was correct except at the end where we indexed the data.frame with "n_opt_vect[i]", which in the case of a numeric vector selected an incorrect column. So we just needed to pass the option name. The error is my fault during the refactoring. 

The reason our test passed previously was completely accidental - the column indexed in the oberauer dataset just happened to be a numeric one, whereas in chenyu's dataset it was a character. I changed the test to make sure it would have failed before the fix. Some other minor cleanup and vectorized the for loops over option columns. I ran chenyu's model and it runs the same after my changes as with after yours.

GidonFrischkorn:
Thanks for your review and additional changes. When running a final test with Chenyu's model. I ran into another issue when printing the results.

image

I have tried other example with providing an integer vector to the `num_options` and there it does not seem to lead to any problems when printing the summary.

We might want to open another issue for this to keep things separated, but I wanted to raise this here, before merging the branch.

GidonFrischkorn:
Sorry, I found the issue. I ran the models with setting `warmup = 2000` and `iter = 2000` resulting in no samples post-warmup. So this was a somewhat stupid mistake on my part -.-

venpopov:
Oh Ive done that too before :D

PR #261, opened 2025-02-13
Add info for posterior predictive plot to m3 article

GidonFrischkorn:
#### Summary
- extend article on `m3` to include information on posterior predictive plots

#### Tests

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -51,7 +51,8 @@ Suggests:
     tidybayes,
     tidyr,
     usethis,
-    waldo
+    waldo,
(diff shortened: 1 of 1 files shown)


venpopov:
Awesome, looks great, thanks! I just added the gghalves package to the Suggests filed. I got an error when rendering the vignette because I don't have that package. When the package is in Suggests, running "devtools::install_dev_deps()" will always install all suggest packages making sure any of use can run vignettes without having to figure out what packages to install.

venpopov (approved):

PR #262, opened 2025-02-13
Change `softmax` behavior

### LINKED ISSUE #254: [feature-request] change softmax function behavior to align with stan
We have the following definition of the softmax in the package:

The softmax function is a bijective function that maps a real vector with length `m-1` to a probability vector with length `m` with all non-zero probabilities.  The present functions define the softmax function and its inverse, both with a tuning parameter.

$$
\Large P(\eta_i) = \frac{e^{\lambda \eta_i}}{1+ \sum_{j=1}^{m-1} e^{\lambda \eta_j}}
$$

It implicitly adds a baseline option. I don't even remember why we coded it like this but it always annoyed me that it doesn't match what I more commonly see: 

$$
\Large P(\eta_i) = \frac{e^{\lambda \eta_i}}{\sum_{j=1}^m e^{\lambda \eta_j}}
$$

The reason I was reminded of this now is that the latter is also how the softmax is defined in Stan (and most places actually):
https://mc-stan.org/docs/functions-reference/matrix_operations.html#softmax-1

I suggest we change it

GidonFrischkorn:
#### Summary
- change the behavior of the `softmax` and `softmaxinv` function to match the Stan documentation

#### Tests

#### Release notes
- the `softmax` function no longer calculates `m` probabilities form a vector of `m-1` weights,  but transforms the weights passed to it to a probability vector of equal length.


venpopov:
I know I opened this issue but I just realized that softmax and softmaxinv are exported functions so we should not change their behavior lest we break someone's code. Also we don't use them anywhere in our code. I suggest we add a deprecated warning for now that they might change at some point and should not be used.

I've actually been thinking of putting together a small package that just mirrors all stan math functions in R for convenience, because there are now a lot of times I've wished I can reproduce some stan code in R, but that's a separate issue.

venpopov (approved):
Tests added and all passing

Issue #269, opened 2025-02-19
[Feature request] Check whether the choice_rule argument is a character


chenyu-psy:
When a factor value is entered into the `choice_rule` argument in `m3`, the Stan code may be generated incorrectly. 

For instance, when I set `choice_rule = factor("softmax", level = c("simple", "softmax"))`, the model unexpectedly uses "simple" as the value instead of "softmax," resulting in incorrect Stan code generation.

![Image](https://github.com/user-attachments/assets/9f97a704-f4a1-41f9-9e20-a4e0a29409ba)

![Image](https://github.com/user-attachments/assets/b51d7280-3b21-4f9b-8eea-77bd55eb0948)

**Why a factor would be enter to the `choice_rule` argument**:
I use the `expand.grid` function to create a data.frame that lists all the models to be conducted. Each column in the data.frame represents a factor. Therefore, when I retrieve parameter values from each column, the values are factor values rather than character values.

**Solution**:
Maybe we can use the `as.character` function to convert any input value to a character to avoid this issue.

GidonFrischkorn:
I will see if I can have a look what happens when you pass a factor. 

Can you share the `bmmformula` that you are using. Reading the Stan code makes it a bit challenging so see how the formula looks and gets converted.

venpopov:
The reason for this problem is that a factor variable in many cases is treat by R as an integer rather than as the character label. In particular, we use a switch statement that expect a character "softmax" or "simple", and it doesn't play nice with factors. It also warns you immediately that this is not normal (using the vignette example):


``` r
library(bmm)

# pass all formulas in one call
full_formula <- bmf(
  corr ~ b + a + c,
  other ~ b + a,
  dist ~ b + d,
  npl ~ b,
  c ~ 1 + cond + (1 + cond || ID),
  a ~ 1 + cond + (1 + cond || ID),
  d ~ 1 + (1 || ID)
)


rule_factor <- factor("softmax", level = c("simple", "softmax"))
str(rule_factor)
#>  Factor w/ 2 levels "simple","softmax": 2

my_model <- m3(
  resp_cats = c("corr", "other", "dist", "npl"),
  num_options = c("n_corr", "n_other", "n_dist", "n_npl"),
  choice_rule = rule_factor,
  version = "custom",
  links = list(c = "log", a = "log", d = "log")
)

code <- stancode(full_formula, oberauer_lewandowsky_2019_e1, my_model)
[...]

PR #270, opened 2025-02-20
Convert `choice_rule` to `character`

GidonFrischkorn:
- Add converstion to `character` for `choice_rule` argument

#### Tests


venpopov (approved):
I’ll merge this now but as I noted in https://github.com/venpopov/bmm/issues/269#issuecomment-2671850544 I’d like to do this more systematically across all functions

PR #287, opened 2025-09-18
286 helpers extract stancode

### LINKED ISSUE #286: [feature-request] Add helpers to extract information from `stancode`
For solving #285 I have written some helper functions that extract the different program blocks of the `stancode` generated for a `bmmodel`

To keep code review manageable I have separated this into its own issue.

GidonFrischkorn:
#### Summary
- add helper functions `extract_stan_blocks` and `extract_parameter_dimensions` to extract information from `stancode` generated by `bmm` and `brms`
- add test to ensure that helper function work properly

#### Tests

#### Release notes
- new helper functions `extract_stan_blocks` and `extract_parameter_dimensions` to extract information from `stancode` generated by `bmm` and `brms`


venpopov:
There is a small issues. You extract blocks based on their order - i.e. the end of block "parameters { ... }" is "transformed parameters". If the order in the stancode doesn't match the code expectations, it fails. Here's an example where I moved the "transformed parameters" block higher up. We shouldn't depend on the order as it might change if brms changes something and `stan` itself doesn't care about order, so a cosmetic change is not unlikely. Also we won't be able to use this function on other custom code if we want to. I will think about how to solve this, but wanted to note the reason.

``` r
stan_code <- "functions {
}
data {
  int<lower=1> N;  // total number of observations
  vector[N] Y;  // response variable
  int prior_only;  // should the likelihood be ignored?
}
transformed parameters {
  real lprior = 0;  // prior contributions to the log posterior
  lprior += student_t_lpdf(Intercept | 3, 0.1, 2.5);
  lprior += student_t_lpdf(sigma | 3, 0, 2.5)
    - 1 * student_t_lccdf(0 | 3, 0, 2.5);
}
transformed data {
}
parameters {
  real Intercept;  // temporary intercept for centered predictors
  real<lower=0> sigma;  // dispersion parameter
}
model {
  // likelihood including constants
  if (!prior_only) {
    // initialize linear predictor term
    vector[N] mu = rep_vector(0.0, N);
    mu += Intercept;
    target += normal_lpdf(Y | mu, sigma);
  }
  // priors including constants
  target += lprior;
}
generated quantities {
  // actual population-level intercept
  real b_Intercept = Intercept;
}"
extracted_program_blocks <- bmm::extract_stan_blocks(stan_code)

extracted_program_blocks$parameters
#> [1] ""
```

<sup>Created on 2026-01-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

GidonFrischkorn:
Good point. I used the function only with stanched generated by brms for which the code blocks are always in the same order. But I agree that we should generalize this behavior so that stancode with a different order of blocks does not break the function.

venpopov:
I fixed that by using brace-matching that finds the appropriate closing brace even when nested internal braces are present in the block. Having all of your tests was really useful to make sure that the new function works the same! I'll look over now the parameter functions

venpopov:
The rest looks good, I'll merge it now

venpopov (approved):

GidonFrischkorn:
Nice :-)

PR #289, opened 2025-09-18
Add `link_transform` function + tests

### LINKED ISSUE #288: [feature-request] utility function to apply links and inverse links to parameters
I suggest that we add a utility function `link_transform` similar to the `brms` functions `link` and `inv_link` to easy the transformation of parameters with specific links.

If you prefer, we could use the `brms` functions, but I thought it might be easier to code these functions ourselves and then have more flexibility in adding links that we might need down the line.

GidonFrischkorn:
#### Summary
- new function `link_transform` to transform parameters according to `links` passed to the function.
- also allows to use `inverse` links

#### Tests

#### Release notes
- new function `link_transform` to transform parameters from native to parameter scale and back according to `links` passed to the function.


venpopov (approved):
Looks good, only minor changes

Issue #281, opened 2025-06-19
[new model] Add `ezdm` estimated with aggregated data


GidonFrischkorn:
Complementing #64 that provides an extended implementation of the `ezdm_robust` that estimates RTs as a mixture of an inverse Gaussian and uniform distributions to account for potential contaminants, I would like to implement a Bayesian hierarchical implementation of the `ezdm` that estimates DDM parameters from aggregated data, that is `mean_rt`, `var_rt`, `n_correct`, and `n_trials`. This builds on work by Adriana F. Chávez De la Peña and Joachim Vandekerckhove: https://osf.io/yg9b5_v1/

@chenyu-psy is actually already working with the `brms` implementation of the likelihood and is estimating parameter recovery for this implementation.

Ideally, I would like to add a `3-par` and `4-par` version, the `4-par` version would allow to also estimate the starting point `zr` by separating `mean_rt` and `var_rt` for responses to the upper and lower bound respectively. I just have to think how to best implement this without making the interface to complicated.

chenyu-psy:
I think the easiest way is to provide an aggregation function to users. Users only need to prepare the non-aggregated data and specify whether they want to implement the `3-par` version or the `4-par` version. The aggregated data can then be passed to `bmm` directly.

GidonFrischkorn:
I have thought about something along those lines, but this function would also have to parse the formula and aggregate the data accordingly. In principle, I like the idea, but it is also additional work and might require quite a bit of additional testing. In particular if we want to keep `bmm` free of `tidyverse` dependency, which I think will be both @venpopov and my preference. I will think about this.

For the start I will more likely provide some documentation in an article for the `bmm` webpage that explains how data should be aggregated for the `ezdm`.

Issue #280, opened 2025-06-19
[new-model] Implement full ddm


GidonFrischkorn:
Based on the likelihood function available in STAN (v. 2.35) I would like to write a wrapper to make the specification of the full DDM a little easier. Likely this model will not run very fast, but this way we have an implementation of a DDM with support of all trial-to-trial variability parameters in the package. And the coding work is minimal.

This is thought to extent the other evidence accumulation models I am currently working on (see #277 and #281)

I like the idea of having different implementations that provide fast algorithms to fit simplified accumulations models (e.g. `ezdm` or `cswald`) but also have the functionality to estimate the full DDM if necessary.

Steps for implementation:

- [x] fill in `model_info`
- [x] implement STAN functions
- [x] specify `check_data.ddm`
- [x] specify `check_model.ddm`
- [x] specify `bmf2bf.ddm` if necessary
- [x] specify `check_formula.ddm`
- [x] specify `configure_model.ddm`
- [x] add function to generate initial values to allow for proper start of sampling
- [x] specify `distribution` functions for the `ddm` model
- [ ] check that `pp_check` and `bridgesampling` works for `cswald`
- [ ] if possible: optimize sampling and speed for model estimation

GidonFrischkorn:
I have added the `ddm` in some commits to the associated branch. And the generation of the initial values seems to work well with the `create_initfun` method I have added on #290

GidonFrischkorn:
While sampling starts fine with `cmdstanr` as a backend, sampling with `rstan` throws an error because of a dimensions mismatch of the expected and passed initial values. For now, I do not really understand where the problem is, as the `stancode` generated for the models and thus also the parameters are the same irrespective of the backend.

So, I suspect that `rstan` somehow treats the initial values differently.

GidonFrischkorn:
I just tested and the problem is model independent. That is for the `sdm` initial values also only work with `cmdstanr` as backend, but `rstan` declares a similar dimensions mismatch as for the `ddm`.

Issue #277, opened 2025-05-19
[new-model] Implement censored shifted Wald Model


GidonFrischkorn:
I want to add the censored shifted Wald model for modeling reaction times and proportion correct in tasks with few errors. In such task the DDM is typically not well identified as the distribution for error responses has few observations which leads to problems with accurately estimating the drift rate and boundary separation.

The model is described by Miller et al (2018): https://journals.sagepub.com/doi/10.1177/0146621617710465

And there is an existing implementation for STAN by Tom Faulkenberry: https://github.com/tomfaulkenberry/censoredShiftedWald

My goal is to implement the STAN functions with a parametrization in terms of cognitive processes: that ist `drift`, `boundary`, and `ndt`. For this, I should be able to reuse most of the code by Tom Faulkenberry.

If you @venpopov have any suggestions for the implementation feel free to comment, here.

PR #290, opened 2025-09-19
285 create initfun

### LINKED ISSUE #285: [feature-request] Add function `create_initfun.bmmodel` for use of `bmmodels` that require initial values to start sampling
To solve our "initial value problem" I would like to add a function that adaptively creates an initial value function based on the `brmsformula` and the stanched generated by a `bmm` call.

I propose to add a property `init_ranges` to all `bmmodel` objects that would benefit from setting initial values for starting the sampling process and access this information to set initial values accordingly.

The biggest challenges I anticipate are:

1. a complex `bmmformula` that suppresses the intercept and includes a combination of effects: e.g. `parameter ~ 0 + group + group:condition + (0 + group:condition | gr(ID, by = group))`
2. non-linear formulas in the `bmmformula` that require to assign which newly generated parameters are linked to which model parameter. Although this issue might be more complex as non-linear formula could also include multiplicative combinations of two non-linear parameters for which initial values can not be reasonably set as a default 

With respect to 2., we might consider throwing a warning that users have to specify initial values by themselves for such non-linear formulas.

GidonFrischkorn:
#### Summary
- add method `create_initfun` to take care of initial values for `bmmodels` that benefit from or require  initial values for sampling
- as an example I have set initial values for the `sdm`
- currently initial values are only set for intercept parameters; for effects they are sampled from a uniform(-0.1,0.1), for random effects (that is `Z` parameters) they are sampled from a uniform(-.5,.5), and covariance matrices / cholesky factors are initialized as diagonal matrices
- for now, I have implemented only minimal tests, as the setup to access the `create_initfun` is somewhat cumbersome

#### Tests

#### Release notes
- new method `create_initfun` for `bmmodels` that require initial values for optimal sampling


GidonFrischkorn:
I think https://github.com/venpopov/bmm/pull/290/commits/4dd618cc8cc238625081d58178fe0398e395c61d has resolved the issue with using the `create_initfun` method with `rstan`. It seems that sometimes a single value is interpreted as a real, but the way some `sd` parameters are declared by `brms` they are set up as vectors.

I found a solution that suggested wrapping the single value in an array explicitly declaring its `dim`.

For now that seems to have done the trick.

GidonFrischkorn:
Short update:

The `create_initfun` method works well with all three evidence accumulation models (`ddm`, `ezdm`, and `cswald`) that are currently under development (#280, #277, #281 ). In addition, I have also further tested that initial values are now working also with `rstan` as a backend.

I have still not come up with a good idea for tests that does not require to actually fit models. So, if you have an idea I would be happy about your input.

R/helpers-inits.R
```diff
@@ -0,0 +1,207 @@
...
+#' @export
+#' @keywords internal developer
+create_initfun <- function(model, data, formula) {
+  UseMethod("create_initfun")
+}
+
+#' @export
+create_initfun.bmmodel <- function(model, data, formula) {
+  # skip to default method if model object contains no init_ranges
+  # somehow this does not seem to work... only if I do this at the bottom
```
venpopov: This is because NextMethod is a regular function - whatever is returned by the nextmethod will be returned here. You do not assign the result to anything and the code continues to execute the remaining lines. It's like calling:

```R
if (is.null(model$init_ranges))  {
   create_initfun.sdm(...)
}
# ... rest of the code
```
When NextMethod is at the end, it returns because R implicitly returns the last evaluated argument
So if you want to skip everything in this function, you can explicitly make an early return:

```R
if (is.null(model$init_ranges)) {
    return(NextMethod("create_initfun"))
  }
```

venpopov:
I did some major refactoring to simplify the multiple nested conditionals and eliminate redunancies. All the tests pass, but please check it against your in progress models, in case something I changed was not caught by the tests. Also, see comment above about the method routing logic

R/helpers-inits.R
```diff
@@ -0,0 +1,156 @@
...
+    if (is.numeric(x)) {
+      if (numeric_strategy == "one") {
+        return(1L) # treat continuous predictor as 1 level
+      } else if (numeric_strategy == "unique") {
+        ux <- if (na.rm) unique(na.omit(x)) else unique(x)
+        return(length(ux))
+      } else {
+        stopif(is.null(nbins) || nbins < 1, "Provide a positive 'nbins' for numeric_strategy='bins'.")
+        x2 <- if (na.rm) x[!is.na(x)] else x
+        if (!length(x2)) {
```
venpopov: I already simplified the function but is this remaining complexity necessary? You only use the count_term_levels function once with the default. Are there any cases when continuous predictors having bins or multiple terms, unless they are already coded as factors?

venpopov (changes requested):

R/helpers-inits.R
venpopov: which in this case I even simplified to directly return 1 instead of calling a method to return 1. Let me know if there was a different reason to call NextMethod?

Currently it doesn't need to be a method because it only depends on whether the model has init_ranges specified, but not what the model is. If you had in mind that different models might have different create_initfun.* methods, then there will need to be a different logic to route it

GidonFrischkorn:
> I did some major refactoring to simplify the multiple nested conditionals and eliminate redunancies. All the tests pass, but please check it against your in progress models, in case something I changed was not caught by the tests.

This looks a lot simpler and a quick test with the current `cswald` implementation did run. The checks with correlations between parameters also ran fine. So from my perspective this should be good for a merge to `develop`.

Have you considered which types of additional tests we could implement to ensure proper functionality?

R/helpers-inits.R
GidonFrischkorn: I have to check this more closely to remind myself why I implemented it this way. Would you prefer resolving this before accepting the pull request? Then I will try to have a look at it tomorrow.

GidonFrischkorn:
I have simplified the helper function and agree that typically numerical predictors should always be handled as factors. I have also reworked one of the tests that was skipped before because it was not properly specified.

venpopov (approved):
Looks good now, thanks. In the future we have see how if at all this needs to be adapted for different models - see my comment above. Currently there is no need for it to be an S3 method, but it's fine to keep it that way

PR #294, opened 2026-01-21
add instructions for coding agents

### LINKED ISSUE #293: [feature-request] add AGENTS.md instructions for AI assistance
Since we chatted about AI use today, I was thinking to formalize some of it. There is a wide standard in creating a file in root directories of projects called [AGENTS.md](https://agents.md/), which is supported by most coding assist tools including [Github Copilot
](https://github.blog/changelog/2025-08-28-copilot-coding-agent-now-supports-agents-md-custom-instructions/). It can contain some general instructions about the coding style, how to run tests, organization of the repository - pretty similar to our developer guide.

venpopov:
Only adds a couple of md files for coding agents.

I wrote a first draft with everything I could think of. Then to my surprise vscode copilot detected it and ran a default prompt to scan the repository and suggested an improved version, which is the one here.

Closes #293


GidonFrischkorn:
This looks great. The only issue I see is that  `AGENTS.md` and `copilot-instructions.md` contain the same information and we should keep them synchronized. Do you see any need for having the duplicate information?

And if so, is there a possibility to keep them synchronized. One option could be to provide a reference in `copilot-instructions.md` to `AGENTS.md` to avoid the redundancy.

Apart from that I think this is a great addition to get more coherent suggestions from AI tools that support us with coding.

venpopov:
> Do you see any need for having the duplicate information?

Not at all and I'm glad you flagged that. When copilot prompted itself to improve my project instructions it created the improved version in the .github folder, which I then copied to Agents.md because that is seen by any ai assistant. I wasn't sure if copilot itself needs the filee in .github. After looking into it now, I see that this was githubs original format but now they support both. So I'll remove the one in .github and merge

PR #296, opened 2026-01-23
Fix #278: Add comma between duplicate parameters in error message

### LINKED ISSUE #278: Enhancement of the error msg.
It is just a small issue, but when the same error occurs twice in the `bmf` formula, there should be a comma added between the error messages.

```
Error: Duplicate formula for parameter(s) m1Duplicate formula for parameter(s) m2
```

venpopov:
## Summary
Fixes #278 

## Changes
- Fixed duplicate formula error message to properly separate parameter names with commas
- Added `duplicates` variable to identify duplicated parameters before using in error message
- Used `collapse_comma()` for consistent formatting (matching pattern in line 127)
- Added test cases for both single and multiple duplicate parameters

## Before
```
Error: Duplicate formula for parameter(s) m1Duplicate formula for parameter(s) m2
```

## After
```
Error: Duplicate formula for parameter(s) 'm1', 'm2'
```

## Testing
- Added new test `test_that("bmmformula gives proper error for duplicate parameters")`
- All existing tests pass
- R CMD CHECK passes with 0 errors, 0 warnings

### DIFF
--- a/R/bmmformula.R
+++ b/R/bmmformula.R
@@ -85,7 +85,8 @@ bmmformula <- function(...) {
   })
 
   stopif(any(vapply(par_names, length, integer(1)) == 0), "Formulas must have a left-hand-side variable")
-  stopif(any(duplicated(par_names)), "Duplicate formula for parameter(s) {par_names[duplicates]}")
+  duplicates <- par_names[duplicated(par_names)]
+  stopif(length(duplicates) > 0, "Duplicate formula for parameter(s) {collapse_comma(duplicates)}")
   new_bmmformula(setNames(out, par_names))
 }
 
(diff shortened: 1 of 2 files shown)


venpopov:
I was testing the new coding agent workflow. I gave it only the following instruction: 

```text
Use gh to see issue #278. Then investigate the issue, make a new branch, fix and open a PR
```

This is the result. The issue was very simple, but it correctly followed our rules and workflow, as well as the use of custom functions and testing rules.

Since it is super simple, I'll merge after the CI check.

venpopov:
I see github also offers to assign copilot for review. I'll do it just as an experiment.

GidonFrischkorn (approved):
Good catch. This is a great addition to avoid conflicting formulas.

Issue #291, opened 2025-09-21
[feature-request] add `ezdm_summary_stats` function for robust summary stats when fitting the `ezdm`


GidonFrischkorn:
There is some evidence that reaction time outliers and random guesses can severely influence parameter recovery of DDM parameters estimated with the EZ-Diffusion Model.

I suggest that we implement a function `ezdm_summary_stats` that estimates robust summary stats, by fitting the mixture of a uniform and a parametric distribution resembling the RT distribution (lognormal, inv_gaussian, or exgaussian) to the data and calculating the moments of the RT distribution based on the parameters of the parametric distribution.

This is an approach that has been suggested in the context of RT modeling quite some time ago. And I think it would provide a considerable improvement in parameter recovery from the `ezdm`.

PR #298, opened 2026-01-23
281 ezdm aggregated data

### LINKED ISSUE #281: [new model] Add `ezdm` estimated with aggregated data

GidonFrischkorn:
#### Summary
- Implements the `ezdm` as a new `bmmodel`. This model uses aggregate statistics and builds them into a probabilistic models based on the sampling distribution of the summary statistics.
- Two versions `3par` and `4par` 
- This pull request is complemented with new features on #291 that provide robust summary statistics to handle contaminant responses. #291 also contains documentation of the `ezdm`  

#### Tests


R/model_ezdm.R
```diff
@@ -0,0 +1,545 @@
...
+.ezdm_version_table <- list(
+  "3par" = list(
+    parameters = list(
+      drift = "Drift rate = Average rate of evidence accumulation of the decision processes",
+      bound = "Boundary separation = Distance between the decision boundaries that need to be reached",
+      ndt = "Non-decision time = Additional time required beyond the evidence accumulation process",
+      s = "The diffusion constant, that is the standard deviation of the Gaussian noise during sampling"
+    ),
+    links = list(
+      drift = "log", bound = "log", ndt = "log", s = "log"
```
venpopov, line 14: Is the log link for drift correct? I think it can take both positive and negative values, right?

GidonFrischkorn: Theoretically yes, they can be negative and positive. There are some issues that I needed to resolve for making the models run stable with an identity link. Just using an identity link leads to discontinuities at zero which led to serious sampling issues. I have created a new branch ezdm_negative_drift that addresses this problem. 

However this required some re-writing of the STAN likelihood functions and thus I thought it would be best to handle this in a separate pull request.

R/model_ezdm.R
```diff
@@ -0,0 +1,545 @@
...
+  mean_rt <- model$resp_vars$mean_rt
+  var_rt <- model$resp_vars$var_rt
+  n_upper <- model$resp_vars$n_upper
+  n_trials <- model$other_vars$n_trials
+
+  # check that required model arguments were provided
+  stopif(
+    is.null(mean_rt) || is.null(var_rt) || is.null(n_upper) || is.null(n_trials),
+    "ezdm model requires mean_rt, var_rt, n_upper, and n_trials arguments"
+  )
```
venpopov: doesn't hurt but unnecessary because the user facing function ezdm already checks for missing arguments with the `stop_missing_args` function. It won't stop if one of the arguments is NULL, but for them to be NULL the user explicitly has to set it to be NULL, which is unlikely. We don't check for this in the other check_data functions.

R/model_ezdm.R
venpopov: ok, I'll trust you on this. But then we should do that before releasing the model as I think it's a typical expectation that drift can be either  (and in the recent paper they do recovery simulations with a full range of both positive an negative drifts)

R/model_ezdm.R
```diff
@@ -0,0 +1,545 @@
...
+  formula <- bmf2bf(model, formula)
+  links <- model$links
+
+  # construct the family & add to formula object
+  ezdm_3par_family <- function(link_drift, link_bound, link_ndt, link_s) {
+    brms::custom_family(
+      'ezdm_3par',
+      dpars = c("mu","drift","bound","ndt","s"),
+      links = c("identity",link_drift,link_bound,link_ndt,link_s),
+      lb = c(NA,0,0,0,0), # lower bounds for parameters
```
venpopov: doesn't need to be a function, as you use it immediately and doesn't add anything

R/model_ezdm.R
```diff
@@ -0,0 +1,545 @@
...
+  )
+}
+
+# posterior_predict for pp_check
+# Returns predictions for EZDM 3par dependent variables
+# By default returns mean_rt (the primary response Y)
+# Use dv argument to select other variables: "var_rt", "n_upper"
+# Usage: posterior_predict(fit, dv = "var_rt")
+posterior_predict_ezdm_3par <- function(i, prep, ...) {
+  dots <- list(...)
```
venpopov: better as an explicit argument that through dots

R/model_ezdm.R
```diff
@@ -0,0 +1,545 @@
...
+configure_model.ezdm_3par <- function(model, data, formula) {
+  # construct brms formula from the bmm formula
+  formula <- bmf2bf(model, formula)
+  links <- model$links
+
+  # construct the family & add to formula object
+  ezdm_3par_family <- function(link_drift, link_bound, link_ndt, link_s) {
+    brms::custom_family(
+      'ezdm_3par',
+      dpars = c("mu","drift","bound","ndt","s"),
```
venpopov: Having the mu parameter here and not being fixed leads to the model pointlessly sampling from it. For the example in the documentation, you get:

```
brms::fixef(fit)
                  Estimate  Est.Error       Q2.5      Q97.5
Intercept        0.6102652 3.97163430 -6.9997988  8.5386128
drift_Intercept  0.6769579 0.03203988  0.6112267  0.7378830
bound_Intercept  0.4020760 0.02679280  0.3494911  0.4540454
ndt_Intercept   -1.2239912 0.03863077 -1.3024472 -1.1513580
s_Intercept      0.0000000 0.00000000  0.0000000  0.0000000
```

I think we either:
- treat the drift as mu and then rename it
- fix mu to be 0

venpopov: add `mu = 0` as a fixed parameter to the models seems to work fine and is much simpler than dealing with the renaming. Now:

```
brms::fixef(fit)
                  Estimate  Est.Error       Q2.5      Q97.5
Intercept        0.0000000 0.00000000  0.0000000  0.0000000
drift_Intercept  0.7171927 0.03425566  0.6467707  0.7823475
bound_Intercept  0.4218990 0.02898746  0.3652506  0.4788026
ndt_Intercept   -1.1973024 0.03689402 -1.2739100 -1.1283095
s_Intercept      0.0000000 0.00000000  0.0000000  0.0000000
```

R/model_ezdm.R
```diff
@@ -0,0 +1,545 @@
...
+  ndt <- brms::get_dpar(prep, "ndt", i = i)
+  s <- brms::get_dpar(prep, "s", i = i)
+
+  # extract n_trials for this observation
+  n_trials <- prep$data$trials[i]
+
+  # number of posterior draws
+  n_draws <- length(drift)
+
+  # generate predictions for each posterior draw
```
venpopov: The rezdm should be able to handle vectorized inputs just fine from my testing, so the sapply is not needed. With the vectorized version you get posterior predictions much much faster (basically instantly)

venpopov (approved):

venpopov:
I made some relatively minor changes. All of the comments above are something I implemented directly, so you don't need to change anything in response. I'll wait on the other PRs until we merge this (feel free to do so if everything seems fine)

venpopov:
I forgot to ask. Are the default priors the same as those used by Adriana and Joachim, which they attribute to a survey paper on ddm?

GidonFrischkorn:
> I forgot to ask. Are the default priors the same as those used by Adriana and Joachim, which they attribute to a survey paper on ddm?

The priors are not the same. Adriana and Joachim use truncated normal distributions and identity link functions for all parameters. This is not possible for us building on `brms`. The priors are however in a similar range. But, I have to check for the effects.

The standard deviation priors for random effects are the default `brms` priors. Which is something we might want to change eventually, as these are very wide for most of the models, especially as many of the parameters use link functions.

venpopov:
> > I forgot to ask. Are the default priors the same as those used by Adriana and Joachim, which they attribute to a survey paper on ddm?
> 
> The priors are not the same. Adriana and Joachim use truncated normal distributions and identity link functions for all parameters. This is not possible for us building on `brms`. The priors are however in a similar range. But, I have to check for the effects.
> 
> The standard deviation priors for random effects are the default `brms` priors. Which is something we might want to change eventually, as these are very wide for most of the models, especially as many of the parameters use link functions.

Makes sense. I just saw that in the paper they mention basing priors on existing literature and I thought that it is nice that this information is available.

PR #299, opened 2026-01-23
291  ezdm summary stats function

### LINKED ISSUE #291: [feature-request] add `ezdm_summary_stats` function for robust summary stats when fitting the `ezdm`

GidonFrischkorn:
#### Summary
- Add `ezdm_summary_stats` function to provide an easy and accessible way to compute robust summary statistics for the `ezdm` model
- Options are `simple` for traditional summary stats, `robust` for median and IQR based summary stats, and `mixture` for summary stats based on a mixture between a uniform and RT-like distribution
- Add documentation for `ezdm` and the use of `ezdm_summary_stats` in an extensive article. This article already includes
#### Tests


R/helpers-data.R
```diff
@@ -301,3 +301,1093 @@ has_nonconsecutive_duplicates <- function(vec) {
...
+    length(contaminant_bound) != 2,
+    "contaminant_bound must be a vector of length 2"
+  )
+  # Validate each element is either numeric (or coercible) or "min"/"max"
+  valid_bound <- function(x) {
+    if (is.numeric(x)) return(TRUE)
+    if (is.character(x)) {
+      # Check if it's "min" or "max"
+      if (tolower(x) %in% c("min", "max")) return(TRUE)
+      # Check if it's a numeric string
```
venpopov: The AI goes overboard with these checks. Something much simpler suffices...

```
  valid_bound <- function(x) {
    is.numeric(x) || tolower(x) %in% c("min", "max")) 
  }
```

R/helpers-data.R
```diff
@@ -301,3 +301,1093 @@ has_nonconsecutive_duplicates <- function(vec) {
...
+      num_val <- suppressWarnings(as.numeric(x))
+      if (!is.na(num_val)) return(TRUE)
+    }
+    FALSE
+  }
+
+  stopif(
+    !valid_bound(contaminant_bound[1]) || !valid_bound(contaminant_bound[2]),
+    "contaminant_bound elements must be numeric or 'min'/'max'"
+  )
```
venpopov, line 459: but this does the same as the more needlessly complex valid_bound function above. I'll simplify all of this but please check these more carefully. We also don't need any of this complexity of checking for numeric bounds encoded as characters, just reject them.

We don't need 30 lines of code to validate the bounds!

venpopov: Also we don't need these endless tests for validating input. This is a bit pointless imo. Some amount of validation and testing when crucial is fine, but we should be testing primarily whether function behavior is correct. I don't want to have to deal with 10 failed tests just for a minor change in the validation logic or message.

R/helpers-data.R
venpopov: oh, I see why are you checking for these numeric strings, to deal with inputs such as c(0.1, "max), which R converts to c("0.1", "max"). I don't like all this complexity. Do we need it?

R/helpers-data.R
```diff
@@ -301,3 +301,1093 @@ has_nonconsecutive_duplicates <- function(vec) {
...
+  tol = 1e-6,
+  adjust_accuracy = FALSE,
+  guess_rate = 0.5,
+  seed = NULL
+) {
+  # Input validation
+  stopif(missing(data), "Argument 'data' is required")
+  stopif(missing(rt), "Argument 'rt' is required")
+  stopif(missing(response), "Argument 'response' is required")
+
```
venpopov, line 433: stop_missing_args() takes care of all these checks

R/helpers-data.R
```diff
@@ -301,3 +301,1093 @@ has_nonconsecutive_duplicates <- function(vec) {
...
+    "distribution must be 'exgaussian', 'lognormal', or 'invgaussian'"
+  )
+  stopif(
+    not_in(method, c("mixture", "simple", "robust")),
+    "method must be 'mixture', 'simple', or 'robust'"
+  )
+  stopif(
+    not_in(robust_scale, c("iqr", "mad")),
+    "robust_scale must be 'iqr' or 'mad'"
+  )
```
venpopov: The standard R way for these checks is to put the vector of possible values in the argument and then just do this:

version <- match.arg(version)

R/helpers-data.R
```diff
@@ -301,3 +301,1093 @@ has_nonconsecutive_duplicates <- function(vec) {
...
+  # Warn if adjust_accuracy is TRUE but method is "simple"
+  warnif(
+    adjust_accuracy && method == "simple",
+    "adjust_accuracy has no effect with method='simple' (no contaminant estimate)"
+  )
+
+  # Set seed for reproducibility if provided
+  if (adjust_accuracy && !is.null(seed)) {
+    set.seed(seed)
+  }
```
venpopov: functions should never use set.seed as it alters the user environment. Either let them set their own seed outside of the function (preferred) or use withr's local seed option which does not affect the global seed

R/helpers-data.R
GidonFrischkorn: There are inconsistent recommendations. And ideally, I would have to explore which settings best recover the proportion of contaminants. From my inital testing, fixed bounds work best if there are no contaminants, because then contaminants are least likely to be overesimated. However, when the data has contaminants, fixed bounds can lead to underestimation of contaminants if not set-properly, so data-driven bounds ("min", "max") tend to be more adequate for detecting contaminants.

If you think, I should take a deeper look into this behavior, I will see when I find time to explore this in depth. There is also some literature checking for the behavior of the different options. For now, I decided to not impose anything by limiting the function, but I also see that this has resulted in a complex and less accessible summary function.

venpopov: Nah, I simpified this without changing the options you have now

GidonFrischkorn: Perfect, thanks for your work!!!

venpopov:
Ok, I'm done. Works nicely, I just had to go through my typical refactoring. I also remove the two functions which you use to convert from a parameter list to a parameter vector and adjusted the code so that optim works. I know why you had it that way and I've ran into the same issue before. But optim can work easily with a named vector, so I just changed all places in the code to work with named parameter vectors instead of lists. It's good to go from my side

venpopov (approved):

PR #300, opened 2026-01-23
Feature/ezdm negative drift

GidonFrischkorn:
#### Summary
- Implementation of an Identity link for drift in the `ezdm` and associated changes to the likelihood functions that allow negative drift for below-chance performance
- Technical details: Soft absolute value ensures MCMC stability
- No breaking changes: Existing code with positive drift still works
- Comprehensive testing: Validation, edge cases, and parameter recovery covered

#### Tests


GidonFrischkorn:
# Soft Absolute Value for Negative Drift Rates

## Problem

The EZDM with an "identity" link drift requires `|drift|` for computing RT moments (MDT, VRT), but need the signed drift for computing proportion correct (pC). When allowing negative drift rates (to model below-chance performance), we need a smooth approximation of the absolute value function that:

1. **Has continuous gradients** (no discontinuity at zero)
2. **Has reasonable curvature** (bounded second derivatives for NUTS sampler)
3. **Approximates** |x| **well enough** for accurate parameter estimation

## The Issue with abs()

Using `abs(drift)` creates a **gradient discontinuity** at drift=0:
- For drift > 0: d|drift|/d(drift) = +1
- For drift < 0: d|drift|/d(drift) = -1
- At drift = 0: gradient is undefined (jumps from -1 to +1)

This breaks Stan's NUTS sampler which requires smooth gradients.

## Solutions

Soft Absolute Value: sqrt(drift² + τ²) with τ = 0.01

image


This pull request, implements this for the `distribution` function and `STAN` chunks & adapts the tests

inst/stan_chunks/ezdm_3par_functions.stan
```diff
@@ -3,6 +3,19 @@ real ezdm_3par_lpdf (real mrt, real mu, real drift, real bound, real ndt, real s
   // Cache common calculations
   real bound_sq = square(bound);
   real s_sq = square(s);
+  
+  // Declare all variables at the top
```
venpopov, line 7: In general can we skip obvious comments? A comment should be needed to explain the reasoning behind a decision or why some piece of code exists, but not explain what the code is when it is straightforward

venpopov (approved):
Looks good, only minor changes

PR #302, opened 2026-01-24
280 full ddm

### LINKED ISSUE #280: [new-model] Implement full ddm

GidonFrischkorn:
#### Summary
- Implements the `ddm` based on the likelihood provided by STAN.
- This provides mostly a convenience implementation to simplify the use of the wiener function available in STAN and with our `create_initfun` the user no longer needs to worry about writing their own initial value function.
- The `7par` version uses a new likelihood function for STAN, but runs very slow. There is some documentation on this in the vignette
- Distribution functions are based on the functions implemented in resists, this is similar to `brms`.

#### Tests


GidonFrischkorn:
The R-CMD-checks fail, because the most recent STAN version is not available here on GitHub. For the 7-parameter wiener likelihood STAN 2.35 or newer is required. Given the feasibility issues with running the 7par version, we could also consider not exposing it.

Maybe, we can have a chat about that.

venpopov:
Judging by the error message it is because the check on github is using rstan as the backend, and rstan is at least a year behind in shipping the STAN  headers. This also means that it will fail on CRAN as well.

There are two options:
- do not include the 7par model
- include it but gate it and its test behind a check if 'cmdstanr" is available. As we do for example in an optional function calc_error_relative_to_nontargets, which checks if tidyr is available:


```R
  stopif(
    !requireNamespace("tidyr", quietly = TRUE),
    'The "tidyr" package is required for this functionality'
  )
```

It's allowed on CRAN to have functions that depend on Suggested packages, but you have to check if they are available. In this case you would check if "cmdstanr" is available and whether the backend is specified as "cmdstanr". The latter is done by default if the package is installed, but users could in principle explicitely override it and request "rstan", in which case you would intentionally give an error message.

GidonFrischkorn:
I have tried a few things to resolve the failing checks. So far none of them worked... Maybe the `ddm` also needs a specific version of `cmdstanr` so that we need to specify that this version is required for the model.

I will check this and get back to you, once I found a solution.

The alternative is to scrap the 7par version completely and basically provide a wrapper around the brms implementation.

GidonFrischkorn:
After some consideration I decided to remove the `7par` version and adjusted the functions accordingly. Now the checks all pass :-)

GidonFrischkorn:
I have decided to evaluate what it would take to implement the mixture between uniform and model likelihood for the rt models directly in STAN for the `ddm` given that the model is practically already available in `brms` and the implementation is mostly improving usability as no initial value function has to be written and the model formula is compiled for the user.

But it seems, that there is more to it, then just adding the mixture in the STAN likelihood. For now, there are some convergence issues. Thus, I switched this pull request to a draft.

GidonFrischkorn:
The implementation of the mixture of uniform and `ddm` distribtuion is more complicated and will require some more comprehensive testing and evaluation of different implementation strategies. So far, I did not manage to find a combination between priors and likelihood specification that results in stable sampling and good model convergence.

I would prefer looking into this for a later release and for now keep the pre-processing with using ML mixtures implemented for #309  as a tool for identifying potential outlier RTs. This is something that work well already and provides an acceptable workflow from my perspective.

So, I re-open this pull request with the `3par` and `4par` version implemented.

GidonFrischkorn:
I have sent you an a self-contained script to verify the DDM implementation via Teams. It compares DDM (trial-level Wiener likelihood) against EZDM (summary statistics) across four scenarios:

1. **Basic hierarchical model** — drift, bound, ndt with random effects (25 subjects, 150 trials)
2. **Condition effects** — easy vs hard drift rate manipulation (75 trials/condition)
3. **Free starting point (zr)** — DDM with `zr ~ 1 + (1|id)` vs EZDM 4par
4. **RT contaminant handling** — 5% injected contaminants, comparing DDM dirty/clean (`flag_contaminant_rts()`) vs EZDM simple/mixture (`ezdm_summary_stats(method="mixture")`)

Expect ~90-120 min for the initial run (the DDM fits take a while). The script prints group-level and subject-level recovery (correlation + 95% CI coverage) and saves comparison plots to the same directory.

venpopov:
@GidonFrischkorn can you resolve the conflicts with the latest branch before I take a look?

GidonFrischkorn:
Sure. I will notify you as soon as I am done with it.

GidonFrischkorn:
@venpopov all conflicts are resolved. `develop` is merged into this branch so there should be no issues merging it once you have reviewed everything.

README.md
venpopov, line 1: The README.md file is generated from README.rmd so these changes will be overwritten and need to be made directly in the rmd file

venpopov (changes requested):
Aside from the readme issue, everything else looks fine

GidonFrischkorn:
Will take care of it asap and then notify you.

venpopov:
After regenerating the readme I noticed that the decision making models had inconsistent categories. Minor fix to that and if the checks all pass, it's good to go

venpopov (approved):

PR #304, opened 2026-01-26
Update `rm3`

### LINKED ISSUE #279: [Improvement] The `rm3` function can be improved to be more convenient
I use this issue to collect all my suggestions for the `rm3` function. New suggestions might be added in the future.

1. It is great that the `rm3` checks whether all the parameters used in the activation functions are provided. However, I don't think it's necessary to provide the `b` parameter, as it is fixed to be 0 by default.
2. It would be more convenient if the `rm3` allowed me to provide the entire formula rather than just the activation formula. If I remember correctly, there should be some inner functions in `bmm` that can extract the activation formulas from the entire formula.
3. I attempted to use the `rm3` function in the `mutate` function. It was not very convenient. I have provided my code below. I wonder if there is any way we can simplify the code.

``` R

data <- data_pars %>%
    mutate(b=0) %>% 
        mutate(
          pars = pmap(select(., all_of(param_names)), 
                      function(...) setNames(c(...), param_names))
        ) %>%
        # Generate the response variables
        mutate(
          results = map(
            pars,
            ~ rm3(
              n = 1,
              size = nResp,
              pars = .x,     
              m3_model = model_m3,
              act_funs = act_form
            )
          )
        ) %>% 
        select(-pars) %>% 
        # Convert each results matrix to a named vector
        mutate(
          results_vec = map(results, ~ setNames(as.numeric(.x[1, ]), colnames(.x)))
        ) %>%
        select(-results) %>%
        # Unnest, using the names from the matrix columns
        unnest_wider(results_vec)
```

GidonFrischkorn:
#### Summary
Addresses GitHub Issue #279 with three major improvements to rm3() and dm3() 
distribution functions for M3 models:

1. **Optional `b` parameter**: The fixed parameter `b` no longer needs to be 
   provided by users - it's automatically added from m3_model$fixed_parameters$b 
   when missing. This simplifies function calls for standard M3 models.

2. **Full formula support**: The act_funs parameter now accepts complete 
   bmmformula objects including both activation formulas and parameter formulas. 
   Activation formulas are automatically extracted based on response categories, 
   allowing users to pass the same formula object used for model fitting.

3. **Automatic unpacking**: New unpack=TRUE option in rm3() returns a named 
   vector when n=1, enabling seamless integration with dplyr::reframe() and 
   tibble::as_tibble_row() for automatic column unpacking in simulation 
   workflows without manual spread operations.

Implementation details:
- Modified .compute_m3_probability_vector() to handle NULL act_funs (deferred 
  evaluation), extract activation functions from full formulas, and auto-add 
  fixed b parameter
- Updated rm3() signature with unpack parameter for convenient rowwise operations
- Changed dm3() and rm3() to use NULL default for act_funs instead of immediate 
  evaluation to prevent errors with custom models
- Added comprehensive documentation with examples for all three improvements
- Updated error messages to provide more informative feedback

Testing:
- Added 8 new tests covering all new functionality
- All 125 distribution tests pass
- Maintains full backward compatibility - no breaking changes

#### Tests

#### Release notes
- Usability improvement of the `rm3` function (fixed parameters do not need to be passed, activation formulas will be automatically extracted, and new option `unpack` for more seamless integration with `tidyverse` functions.

### DIFF
--- a/R/distributions.R
+++ b/R/distributions.R
@@ -572,15 +572,24 @@ rimm <- function(n, mu = c(0, 2, -1.5), dist = c(0, 0.5, 2),
 #'   and each value is the number of observed responses per category
 #' @param n Integer. Number of observations to generate data for
 #' @param size The total number of observations in all categories
-#' @param pars A named vector of parameters of the memory measurement model
+#' @param pars A named vector of parameters of the memory measurement model.
+#'   Note: The fixed parameter `b` does not need to be provided - it will be
+#'   automatically added from the model specification if missing.
 #' @param m3_model A `bmmodel` object specifying the m3 model that densities or
 #'   random samples should be generated for
 #' @param act_funs A `bmmformula` object specifying the activation functions for
-#'   the different response categories for the "custom" version of the M3. The
-#'   default will attempt to construct the standard activation functions for the
-#'   "ss" and "cs" model version. For a custom m3 model you need to specify the
-#'   act_funs argument manually
+#'   the different response categories. This can be either:
+#'   (1) Just the activation formulas (one for each response category), or
+#'   (2) A full bmmformula including both activation formulas and other parameters.
+#'   If a full formula is provided, only the formulas matching response categories
+#'   will be extracted. The default will attempt to construct the standard
+#'   activation functions for the "ss" and "cs" model version. For a custom m3
+#'   model you need to specify the act_funs argument manually.
 #' @param log Logical; if `TRUE` (default), values are returned on the log scale.
+#' @param unpack Logical; if `TRUE` and `n = 1`, returns a named vector instead of
+#'   a matrix. This allows automatic unpacking of response categories into separate
+#'   columns when used with `dplyr::reframe()`. Default is `FALSE` for backward
+#'   compatibility.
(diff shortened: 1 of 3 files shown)


GidonFrischkorn:
@chenyu-psy Maybe you can checkout if these changes meet your requirements. At least, you can now use the function with `reframe`. See example in the new documentation.

venpopov:
should I wait for a response from @chenyu-psy before reviewing it?

GidonFrischkorn:
You can already review it and checkout if there are any issues with respect to the implementation, and if anything needs re-factoring. Then we can still wait and see if Chenyu is satisfied with the usability now.

chenyu-psy (approved):

chenyu-psy:
These improvements look good to me. I initially thought the code in the example that uses `rowwise`, `reframe`, and `as_tibble_row` could be slow (because of the `rowwise` function) and complicated for users to understand. A potential solution could be the `purrr::pmap` and `unnest` functions. However, these functions only improve the speed but not the readability of the code. I think I am good with the current example.

venpopov (approved):

PR #306, opened 2026-01-26
Update `construct_default_priors_list` to accomodate bmmformulas more…

### LINKED ISSUE #305: [bug] Default Priors Not Applied to Interaction Terms in Formula Pattern 0 + A + A:B
**Describe the bug**
When using a formula with a suppressed intercept, a main effect, and explicit interaction terms (e.g., kappa ~ 0 + group + group:cond), bmm fails to set priors on the interaction coefficients. Instead, the interactions receive flat priors despite the model specifying an effects prior of normal(0, 1).

Expected Behavior
For the formula kappa ~ 0 + group + group:cond with a mixture2p model:

Main effect levels (groupA, groupB) should get the main prior: normal(2, 1)
Interaction terms (groupA:condy, groupB:condy) should get the effects prior: normal(0, 1)
This matches the behavior of similar patterns like 0 + group * cond where class-level effects priors correctly apply to interaction terms.

Actual Behavior
Main effect levels get normal(2, 1) ✓
Interaction terms get flat priors ✗
A class-level normal(0, 1) prior is set but gets overridden by explicit empty priors on individual coefficients

**To Reproduce**
```
library(bmm)

data <- oberauer_lin_2017
data$group <- factor(rep(c('A', 'B'), length.out = nrow(data)))
data$cond <- factor(rep(c('x', 'y'), each = nrow(data)/2))

formula <- bmf(kappa ~ 0 + group + group:cond, thetat ~ 1)
model <- mixture2p('dev_rad')

# Check what bmm sets
bmm_prior <- default_prior(formula, data, model)
print(bmm_prior[bmm_prior$dpar == 'kappa', c('prior', 'class', 'coef')])
```

**Expected behavior**
The Interaction terms, should get the default `effects` priors

The cause for this is that the construct_default_priors_list() function has three branches:

Intercept present → sets intercept prior
Single fixed effect OR interaction-only → sets class-level prior
"Edge case" (lines 267-274) → sets coefficient-level priors on first predictor's levels
For 0 + group + group:cond:

fixed_effects_count = 1, interactions_count = 1
Falls to branch 3 (edge case)
Code only sets priors on groupA, groupB
Does NOT set priors on groupA:condy, groupB:condy
Later, brms::default_prior() adds explicit empty priors for ALL coefficients. These explicit empty priors on the interaction terms override the class-level normal(0, 1) effects prior, resulting in flat priors.

GidonFrischkorn:
### Issues
**Bug 1: Multiple predictors without intercept**
Formulas like kappa ~ 0 + set_size + session only set priors on the first predictor (set_size), leaving subsequent predictors with flat priors.

**Bug 2: Interaction-only with intercept**
Formulas like kappa ~ 1 + set_size:session resulted in flat class-level priors for interaction terms instead of inheriting from the effects prior.

### Solution

- Restructured logic to set class-level effects prior only when intercept is present OR for multi-predictor formulas without intercept
- Added condition to check interactions_count > 0 alongside fixed_effects_count > 0 for setting class-level effects priors
- For multi-predictor formulas without intercept (e.g., 0 + A + B), first predictor now gets main prior, remaining predictors get effects prior (matching brms' statistical interpretation)

### Tests
Added comprehensive coverage validating:

- Multi-predictor handling with/without intercept (0 + A + B, 1 + A + B)
- Interaction-only formulas (1 + A:B, 0 + A:B)
- Proper main vs effects prior distinction across patterns
- Invariance to factor contrast coding (contr.treatment, contr.sum, contr.helmert, contr.equalprior)

#### Release notes

Bug fixes

- Fixed default prior specification for formulas with multiple predictors without intercept (e.g., 0 + A + B) - now correctly assigns main prior to first predictor and effects prior to remaining predictors
- Fixed default prior specification for interaction-only formulas with intercept (e.g., 1 + A:B) - now correctly sets class-level effects prior instead of flat priors


venpopov:
Are you sure there is a problem to be solved? When I tested the supposed problem cases I get correct priors on the develop branch:

```r
library(bmm)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
data <- oberauer_lin_2017
data$group <- factor(rep(c('A', 'B'), times = nrow(data)/2))
data$cond <- factor(rep(c('x', 'y'), each = nrow(data)/2))

model <- mixture2p('dev_rad')

# Check what bmm sets
print(filter(default_prior(bmf(kappa ~ 0 + group:cond), data, model), nlpar == 'kappa'))
#> No formula for parameter thetat provided. Only a fixed intercept will be estimated.
#>         prior class         coef group resp dpar nlpar   lb   ub tag
#>  normal(2, 1)     b groupA:condx                 kappa <NA> <NA>    
#>  normal(2, 1)     b groupA:condy                 kappa <NA> <NA>    
#>  normal(2, 1)     b groupB:condx                 kappa <NA> <NA>    
#>  normal(2, 1)     b groupB:condy                 kappa <NA> <NA>    
#>  normal(2, 1)     b                              kappa <NA> <NA>    
#>        source
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>          user
print(filter(default_prior(bmf(kappa ~ 0 + group + cond), data, model), nlpar == 'kappa'))
#> No formula for parameter thetat provided. Only a fixed intercept will be estimated.
#>         prior class   coef group resp dpar nlpar   lb   ub tag       source
#>  normal(0, 1)     b  condy                 kappa <NA> <NA>     (vectorized)
#>  normal(0, 1)     b                        kappa <NA> <NA>             user
#>  normal(2, 1)     b groupA                 kappa <NA> <NA>             user
#>  normal(2, 1)     b groupB                 kappa <NA> <NA>             user
print(filter(default_prior(bmf(kappa ~ 0 + group + group:cond), data, model), nlpar == 'kappa'))
#> No formula for parameter thetat provided. Only a fixed intercept will be estimated.
#>         prior class         coef group resp dpar nlpar   lb   ub tag
#>  normal(0, 1)     b groupA:condy                 kappa <NA> <NA>    
#>  normal(0, 1)     b groupB:condy                 kappa <NA> <NA>    
#>  normal(0, 1)     b                              kappa <NA> <NA>    
#>  normal(2, 1)     b       groupA                 kappa <NA> <NA>    
#>  normal(2, 1)     b       groupB                 kappa <NA> <NA>    
#>        source
#>  (vectorized)
#>  (vectorized)
#>          user
#>          user
#>          user
# Check what bmm sets
print(filter(default_prior(bmf(kappa ~ 0 + set_size:session), data, model), nlpar == 'kappa'))
#> No formula for parameter thetat provided. Only a fixed intercept will be estimated.
#>         prior class               coef group resp dpar nlpar   lb   ub tag
#>  normal(2, 1)     b set_size1:session1                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size1:session2                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size2:session1                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size2:session2                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size3:session1                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size3:session2                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size4:session1                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size4:session2                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size5:session1                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size5:session2                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size6:session1                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size6:session2                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size7:session1                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size7:session2                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size8:session1                 kappa <NA> <NA>    
#>  normal(2, 1)     b set_size8:session2                 kappa <NA> <NA>    
#>  normal(2, 1)     b                                    kappa <NA> <NA>    
#>        source
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>          user
print(filter(default_prior(bmf(kappa ~ 0 + set_size + session), data, model), nlpar == 'kappa'))
#> No formula for parameter thetat provided. Only a fixed intercept will be estimated.
#>         prior class      coef group resp dpar nlpar   lb   ub tag       source
#>  normal(0, 1)     b  session2                 kappa <NA> <NA>     (vectorized)
#>  normal(0, 1)     b                           kappa <NA> <NA>             user
#>  normal(2, 1)     b set_size1                 kappa <NA> <NA>             user
#>  normal(2, 1)     b set_size2                 kappa <NA> <NA>             user
#>  normal(2, 1)     b set_size3                 kappa <NA> <NA>             user
#>  normal(2, 1)     b set_size4                 kappa <NA> <NA>             user
#>  normal(2, 1)     b set_size5                 kappa <NA> <NA>             user
#>  normal(2, 1)     b set_size6                 kappa <NA> <NA>             user
#>  normal(2, 1)     b set_size7                 kappa <NA> <NA>             user
#>  normal(2, 1)     b set_size8                 kappa <NA> <NA>             user
print(filter(default_prior(bmf(kappa ~ 0 + set_size + set_size:session), data, model), nlpar == 'kappa'))
#> No formula for parameter thetat provided. Only a fixed intercept will be estimated.
#>         prior class               coef group resp dpar nlpar   lb   ub tag
#>  normal(0, 1)     b set_size1:session2                 kappa <NA> <NA>    
#>  normal(0, 1)     b set_size2:session2                 kappa <NA> <NA>    
#>  normal(0, 1)     b set_size3:session2                 kappa <NA> <NA>    
#>  normal(0, 1)     b set_size4:session2                 kappa <NA> <NA>    
#>  normal(0, 1)     b set_size5:session2                 kappa <NA> <NA>    
#>  normal(0, 1)     b set_size6:session2                 kappa <NA> <NA>    
#>  normal(0, 1)     b set_size7:session2                 kappa <NA> <NA>    
#>  normal(0, 1)     b set_size8:session2                 kappa <NA> <NA>    
#>  normal(0, 1)     b                                    kappa <NA> <NA>    
#>  normal(2, 1)     b          set_size1                 kappa <NA> <NA>    
#>  normal(2, 1)     b          set_size2                 kappa <NA> <NA>    
#>  normal(2, 1)     b          set_size3                 kappa <NA> <NA>    
#>  normal(2, 1)     b          set_size4                 kappa <NA> <NA>    
#>  normal(2, 1)     b          set_size5                 kappa <NA> <NA>    
#>  normal(2, 1)     b          set_size6                 kappa <NA> <NA>    
#>  normal(2, 1)     b          set_size7                 kappa <NA> <NA>    
#>  normal(2, 1)     b          set_size8                 kappa <NA> <NA>    
#>        source
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>  (vectorized)
#>          user
#>          user
#>          user
#>          user
#>          user
#>          user
#>          user
#>          user
#>          user
```

GidonFrischkorn:
The bugs are real, but your examples don't expose them for two reasons:

1. You only test `kappa`, which already has an `effects` prior on develop. The bug affects parameters without effects priors (like `thetat`).

Compare your single-formula approach vs specifying both parameters:

```r
library(bmm)
library(dplyr)
data <- oberauer_lin_2017
model <- mixture2p("dev_rad")

# Your approach — looks fine because kappa has effects prior defined:
pr <- default_prior(bmf(kappa ~ 0 + set_size + session), data, model)
filter(pr, nlpar == "kappa", coef == "session2")
#>  normal(0, 1)     b session2  kappa  (vectorized)

# But specify thetat with the same formula — thetat gets flat priors:
pr2 <- default_prior(
  bmf(kappa ~ 0 + set_size + session, thetat ~ 0 + set_size + session),
  data, model
)
filter(pr2, coef == "session2")
#>  (flat)     b session2  kappa default
#>  (flat)     b session2 thetat default
```

`thetat` has no effects prior defined on develop, so it gets completely flat priors for all regression coefficients. Same issue applies to `mu1` (mixture models) and `mu` (SDM).

2. You tested kappa ~ 0 + set_size:session (no intercept), but the code bug is in kappa ~ 1 + set_size:session (with intercept).

```
# No intercept + interaction — works fine (your test):
pr <- default_prior(bmf(kappa ~ 0 + set_size:session), data, model)
filter(pr, nlpar == "kappa", coef == "")
#>  normal(2, 1)     b    kappa  user

# WITH intercept + interaction — all interactions get flat priors:
pr <- default_prior(bmf(kappa ~ 1 + set_size:session), data, model)
filter(pr, nlpar == "kappa", class == "b")
#>  (flat)     b                    kappa default
#>  (flat)     b set_size1:session1 kappa default
#>  ...all 16 interaction terms are (flat)...
#>  normal(2, 1)  b Intercept       kappa user
```

The cause: construct_default_priors_list checks fixed_effects_count > 0 to decide whether to set the class-level effects prior, but interactions have order > 1 so they're not counted. The fix changes this to fixed_effects_count > 0 || interactions_count > 0.

venpopov:
I see. There are a couple of things going on here:

1. I tested with kappa because that's what the original issue describes! So there is no bug 1 in constructing the priors - rather just missing effect priors for some parameters like theta. When you include theta yes you don't get effect priors on it but that's not specific to "Multiple predictors without intercept" - just reflects that we hadn't put effect priors on theta.

The fact that it shows also flat for kappa in your example is an artefact of filtering the prior df and how brms deals with this - the prior on kappa effects is only defined in the generic "b" row, and the summary "pretends it is defined for all b rows, but if you filter it it shows the underlying empty field. Without filtering it looks like:

```
          prior     class      coef group resp   dpar  nlpar   lb   ub tag       source
 logistic(0, 1)    theta2                                    -Inf  Inf          default
   normal(0, 1)         b  session2                    kappa <NA> <NA>     (vectorized)
         (flat)         b                             thetat                    default
         (flat)         b  session2                   thetat               (vectorized)
   normal(0, 1)         b                              kappa <NA> <NA>             user
   normal(2, 1)         b set_size1                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size2                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size3                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size4                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size5                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size6                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size7                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size8                    kappa <NA> <NA>             user
```

So yes, missing effects on theta, but not on kappa. Thetat still misses priors even in the simplest formula that has effects:

```
pr2 <- default_prior(
  bmf(kappa ~ 0 + set_size + session, thetat ~ session),
  data, model
)
pr2
---
          prior     class      coef group resp   dpar  nlpar   lb   ub tag       source
 logistic(0, 1)    theta2                                    -Inf  Inf          default
   normal(0, 1)         b  session2                    kappa <NA> <NA>     (vectorized)
         (flat)         b                             thetat                    default
         (flat)         b  session2                   thetat               (vectorized)
   normal(0, 1)         b                              kappa <NA> <NA>             user
   normal(2, 1)         b set_size1                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size2                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size3                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size4                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size5                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size6                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size7                    kappa <NA> <NA>             user
   normal(2, 1)         b set_size8                    kappa <NA> <NA>             user
 logistic(0, 1)         b Intercept                   thetat <NA> <NA>             user
    constant(0) Intercept                         mu1        <NA> <NA>             user
    constant(0) Intercept                         mu2        <NA> <NA>             user
 constant(-100) Intercept                      kappa2        <NA> <NA>             user
```

I'm not saying that we shouldn't add effect priors to the parameters that lack them, but I was confused by the description of the bug because I don't see a bug.

2. I tested `kappa ~ 0 + set_size:session (no intercept)` because `kappa ~ 1 + set_size:session` is not a proper model specification. It includes both an intercept and a separate parameter for every condition combination, which is not identifiable and shouldn't be run at all! Here's the model matrix for such a formula:

``` r
dat <- expand.grid(f1 = letters[1:3], f2 = LETTERS[1:2])
dat$y <- rnorm(nrow(dat), mean = as.numeric(dat$f1) * as.numeric(dat$f2))
model.matrix(y ~ 1 + f1:f2, data = dat)
#>   (Intercept) f1a:f2A f1b:f2A f1c:f2A f1a:f2B f1b:f2B f1c:f2B
#> 1           1       1       0       0       0       0       0
#> 2           1       0       1       0       0       0       0
#> 3           1       0       0       1       0       0       0
#> 4           1       0       0       0       1       0       0
#> 5           1       0       0       0       0       1       0
#> 6           1       0       0       0       0       0       1
#> attr(,"assign")
#> [1] 0 1 1 1 1 1 1
#> attr(,"contrasts")
#> attr(,"contrasts")$f1
#> [1] "contr.treatment"
#> 
#> attr(,"contrasts")$f2
#> [1] "contr.treatment"
```

When you try to estimate such a model, you get a singularity and NA values, and at least with `lm` it arbitrarily chooses the last level to be dropped and implicitly acts as the intercept:

``` r
dat <- expand.grid(id = 1:5, f1 = letters[1:3], f2 = LETTERS[1:2])
dat$y <- rnorm(nrow(dat), mean = as.numeric(dat$f1) * as.numeric(dat$f2))

lm(y ~ 1 + f1:f2, data = dat) |> summary()
#> 
#> Call:
#> lm(formula = y ~ 1 + f1:f2, data = dat)
#> 
#> Residuals:
#>     Min      1Q  Median      3Q     Max 
#> -1.9306 -0.6243 -0.0380  0.5917  3.7008 
#> 
#> Coefficients: (1 not defined because of singularities)
#>             Estimate Std. Error t value Pr(>|t|)    
#> (Intercept)   6.4577     0.5747  11.237 4.81e-11 ***
#> f1a:f2A      -5.8709     0.8127  -7.224 1.83e-07 ***
#> f1b:f2A      -4.0398     0.8127  -4.971 4.48e-05 ***
#> f1c:f2A      -3.5928     0.8127  -4.421 0.000181 ***
#> f1a:f2B      -4.8769     0.8127  -6.001 3.40e-06 ***
#> f1b:f2B      -2.5487     0.8127  -3.136 0.004482 ** 
#> f1c:f2B           NA         NA      NA       NA    
#> ---
#> Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1
#> 
#> Residual standard error: 1.285 on 24 degrees of freedom
#> Multiple R-squared:  0.7257, Adjusted R-squared:  0.6686 
#> F-statistic:  12.7 on 5 and 24 DF,  p-value: 4.245e-06
```

This is degenerate and lm tries to fix it. I don't even know what happens if you try to run it with brms.

venpopov:
I'm sorry that I phrased my original messages so confrontationally. I couldn't sleep and it came off sharper than I intended - I was just trying to understand what the issue is so that I know what the solution is aiming to do to review it properly. I am super grateful for all the work you've done on these PRs - they both extend the modeling options substantially and make the user experience much better and easier.

GidonFrischkorn:
No worries. I totally understand that you want to understand the issue.

I have to think about this a bit and not handle this in between tasks. So, let me get back to you, once I have found time to sort out what you have raised and what my initial concerns were.

I see your point that `1 + A:B` is not appropriate and is an inadequate model specification and thus we should probably not include it as a test case.

Will get back to you ASAP.

GidonFrischkorn:
I finally found some time to organize my thoughts. I am sorry for the confusing initial description. After a thorough analysis, I now see you were largely right, there is really only one bug here.

## The bug: Missing `effects` priors in model definitions

You correctly identified that the original issue description was misleading — it tested `kappa`, which already has an `effects` prior on develop and works fine. The real problem is that several parameters across models never had `effects` priors defined at all:

| Model | Parameter | `effects` on develop |
|-------|-----------|---------------------|
| SDM | `mu` | missing |
| mixture2p | `mu1`, `thetat` | missing |
| mixture3p | `mu1`, `thetat`, `thetant` | missing |
| IMM | `mu1` | missing |

This means **any** formula with regression coefficients on these parameters gets flat priors, even the simplest case:

```r
# On develop:
pr <- default_prior(
  bmf(kappa ~ set_size, thetat ~ set_size),
  oberauer_lin_2017, mixture2p("dev_rad")
)

# kappa class-level b: normal(0, 1)  ✓
# thetat class-level b: (flat)       ✗  ← no effects prior defined
```

The same applies to interaction terms — for `thetat ~ 0 + group + group:set_size`, the interaction coefficients get flat priors. But this isn't a logic bug in `construct_default_priors_list()` — it's simply that thetat has no effects prior to apply. For `kappa` (which does have an effects prior), the class-level prior is correctly set and the interaction terms inherit from it, just as you showed.

The fix adds `effects` priors to all affected parameters (e.g., `normal(0,1)` for location parameters, `normal(0,0.5)` for mixing proportions on logit scale).

## Secondary improvement: `construct_default_priors_list()` restructuring

The PR also restructures the logic in `construct_default_priors_list()` to be cleaner and handle edge cases more robustly. The only case where the old logic actually failed was the degenerate `1 + A:B` pattern (intercept + interaction-only, no main effects), where `fixed_effects_count = 0` caused the effects prior condition to be skipped entirely. Since you correctly pointed out that `1 + A:B` is an unidentifiable specification, I've removed all tests for that pattern.

The restructured code is still an improvement in clarity — it groups all intercept-related logic together rather than splitting it across two separate `if` blocks — but the practical impact is limited to the degenerate case.

If you would rather keep the old specification I am also fine with that.

## Addressed feedback

- **Removed `1 + A:B` test cases** — you're right that it's a degenerate specification. Removed from all three test blocks (fixed effects, interaction-only, and contrast coding tests). The `0 + A:B` tests remain since those are valid.

venpopov:
Thanks for digging deeper! My only concern with the priors of effects is knowing whether those specific ones are a good default. I remember that when we originally built this functionality we only put priors on main effects because it was easier to reason about what happens there and I think we just decided to avoid effect priors for the moment, like brms, to evaluate how they might affect modeling first. I don't mind adding them, but if we do we should make sure that they make sense and don't have strong or biasing effects on modeling outcomes - especially because we claim that when we do put defaults, they are domain informed. It's easier to say "you should take care about your effect priors" to users if the default is flat. Either option is reasonable, but if we do put priors, even if wesk ones, we need to test it first

venpopov:
This is particularly difficult for example for parameters on transformed scales like `mu` for the circular models. For them mu has a tan_half link. With a tan_half link, the current main effect prior student_t(1,0,1) is a uniform prior over the native scale (https://github.com/paul-buerkner/brms/issues/1631). As a result, if the prior for effects $\Delta \mu$ is Normal(0, 1), now these two models have different priors:

```
(1) mu ~ 0 + some_factor
(2) mu ~ some_factor
```

in (1), every level gets the student_t(1,0,1) prior, so they are all uniformly distributed on the native scale
in (2), the intercept level gets the student_t prior, and is uniformly distributed, whereas the effect differences, once added to the intercept, and converted on the native scale, result in non-reference levels looking like this:

``` r
n = 1e6
atan_double <- \(x) 2*atan(x)
(rt(n, 1) + rnorm(n)) |> 
  atan_double() |> 
  hist(, breaks = 60)
```



<sup>Created on 2026-03-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

GidonFrischkorn:
Hmm... I see your point. With the current uniform prior over -pi to pi, the proposed effects prior produces these results because intercepts far away from zero are quite likely.

We could consider using a somewhat more informative prior that introduces some regularization towards zero for `mu`, which would make sense given the coding of the model that `mu` represents the error from the target and we would assume biases to be close to zero.

This would circumvent this issue, right?

Based on some testing I would suggest

```{r}
samples <- 1e6
intercept <- rnorm(samples, sd = 0.5)
effect <- rnorm(samples, sd = 0.25)

atan_double <- \(x) 2*atan(x)

((intercept + effect) - (intercept - effect)) |> 
  atan_double() |> 
  hist(,breaks = 60)
```
image

and with treatment coding
```{r}
samples <- 1e6
intercept <- rnorm(samples, sd = 0.5)
effect <- rnorm(samples, sd = 0.25)

atan_double <- \(x) 2*atan(x)

((intercept + effect) - intercept) |> 
  atan_double() |> 
  hist(,breaks = 60)
```
image

GidonFrischkorn:
While thinking about the topic of default priors and writing the workflow suggestion I posted on teams this morning, I thought we should also discuss a related issue: random effects SD priors.

Currently, bmm doesn't set any custom priors for random effects standard deviations — we rely entirely on the brms default `student_t(3, 0, 2.5)`. This is very wide for our constrained parameters:

- **kappa (log link):** SD = 2.5 on the log scale means individual kappas could vary by a factor of `exp(2.5) ≈ 12×` from the group mean — implying some participants have essentially zero precision while others have extreme precision.
- **thetat (logit scale):** SD = 2.5 means individual mixing weights spanning nearly the full [0, 1] range.
- **mu (tan_half):** SD = 2.5 means individual biases spanning the entire circle.

I think we should consider parameter-type-specific SD priors as a follow-up. This would require extending `set_default_prior()` to handle `class = "sd"` priors — clearly a separate PR. But I wanted to get your thoughts on whether parameter-specific SD priors are worth pursuing, or if you'd prefer to keep this delegated to brms.

venpopov:
@GidonFrischkorn I don't remember where we left this off. Is this still relevant or should we close it?

GidonFrischkorn:
@venpopov I think generally this is still relevant, but I would close this PR and replace it with a much smaller one. You were right on the diagnosis. But  I think we should still address the missing `effects` priors for some models. Scanning `default_priors` across all supported models on develop, 7 of 27 parameter entries have no `effects` prior:

| Model | Parameter |
|---|---|
| imm | `mu1` |
| mixture2p | `mu1`, `thetat` |
| mixture3p | `mu1`, `thetat`, `thetant` |
| sdm | `mu` |

Everything else has one. That is the whole issue.

Your concern about the `tan_half` link for `mu` parameters in the circular models is legitimate, and I also made an error in the diagnostic I posted earlier: `2*atan(i + e) - 2*atan(i)` is not `2*atan(e)`. Both of my histograms plotted the transform of a latent difference rather than the difference of transformed values, so they show the wrong quantity. For the `intercept ~ N(0, 0.5)`, `effect ~ N(0, 0.25)` proposal, the actual native-scale difference has SD 0.42, while my plot showed 0.85. Roughly a factor of two off.

## My suggestion

1. Close this PR.
2. Open a small one adding `effects = "normal(0, 0.5)"` for `mixture2p$thetat` and `mixture3p$thetat`/`thetant` only, with a prior predictive check on the softmax scale for the mixture3p pair, and a NEWS entry. No changes to `helpers-prior.R`.
3. Split `mu`/`mu1` (imm, mixture2p, mixture3p, sdm) into a separate issue. It is entangled with whether we also change the main prior
4. Drop the `construct_default_priors_list()` restructuring entirely. 
5. Add mock-fit regression tests across formula shapes and contrast codings for default priors.

## Still open

The random effects SD question from March is unchanged: `construct_default_priors_list()` reads only `main` and `effects`, no model defines an `sd` entry, and the only `class = "sd"` prior in the package is the set_size-1 constraint helper. So we are still on brms' `student_t(3, 0, 2.5)` everywhere, which is wide for log-scale `kappa` (a factor of ~12 spread) and logit-scale `thetat`.

PR #309, opened 2026-01-28
307 flag contaminant rts

### LINKED ISSUE #307: [feature-request] pre-process rt data to flag contaminants for `cswald` and `ddm`
Similar to contaminant RTs (fast guesses, or slow responses due to lapses in attention) that bias the `ezdm` estimation, these RTs can also bias estimation of `cswald` and `ddm` parameters. 

I would like to implement a function `flag_contaminant_rts` that implements a similar mixture fitting of a uniform and RT like distritbution (lnorm, invgaussian, or exgaussian), but instead of calculating summary statistics like in the `ezdm_summary_statistics` function, the function should estimate the likelihood of being a contaminant RT for each RT obeservation. This can either be a continuous likelihood ratio between the guessing an RT likelihood or a probability or the result of a probabilistic guess based on a binomial sampling for each RT.

This could then be used to filter RTs with a high probability of being contaminant responses and filter them from the analysis. This is a lot faster than implementing the mixture in STAN and also mirrors the pre-processing for the `ezdm`-

GidonFrischkorn:
Implements `flag_contaminant_rts()` for trial-level contamination detection in RT data.
   
## Features

- Identifies fast guesses and attention lapses using mixture modeling
- Supports 3-parameter (pooled) and 4-parameter (boundary-specific) versions
- Returns contamination probabilities, likelihood ratios, or binary flags
- Bayesian validation of fast guess assumptions for 2AFC tasks via validate_fast_guessing parameter
- Exported validate_fast_guesses() function for standalone validation with Beta-Binomial analysis
- Flexible threshold modes: quantile (adaptive) or absolute (fixed RT)
- Fully compatible with existing bmm workflow

## Testing

- 37 tests covering all functionality including validation features
- All 1084 tests passing across package
- R CMD CHECK: 0 errors, 0 warnings

## Documentation

- Complete roxygen2 documentation with examples showing both threshold modes
- Comprehensive vignette (1118 lines) covering:
  - Methodology and validation approach
  - EZDM and DDM/csWald workflows with validation examples
  - Case study with real data
  - Best practices and troubleshooting
- `NEWS.md` entry added

#### Tests


R/helpers-data.R
```diff
@@ -845,3 +844,830 @@ ezdm_summary_stats <- function(
...
+  if (!is.null(seed)) {
+    stopif(
+      !is.numeric(seed) || length(seed) != 1,
+      "seed must be a single numeric value or NULL"
+    )
+  }
+
+  warnif(any(data[[rt]] > 10), "Some RT values > 10. Ensure RTs are in seconds, not milliseconds.")
+  warnif(any(data[[rt]] <= 0, na.rm = TRUE), "Non-positive RT found. These values will be excluded.")
+
```
venpopov: This should be an error, not a warning. Negative values are serious enough data problem that we should let the user handle it

R/helpers-data.R
```diff
@@ -845,3 +844,830 @@ ezdm_summary_stats <- function(
...
+  denom <- numer_c + numer_rt
+
+  contam_prob <- numer_c / denom
+  contam_lr <- numer_c / numer_rt
+
+  # Probabilistic flagging using binomial sampling
+  # Each trial is flagged with probability equal to its contamination probability
+  if (!is.null(seed)) {
+    set.seed(seed)
+  }
```
venpopov: we should never alter the seed from inside of functions. This affects the global state and can affect reproducibility of user's scripts if they use our functions before some other analyses. And you also have to pass around the seed argument from a ton of function arguments to others just to use it once here. We should remove all seed mentions and leave it to the user to control reproducibility

venpopov: I know that for some reason copilot keeps using this anti-pattern in my other projects too. We should explicitly forbid it in the agents.md file

venpopov (changes requested):
There is a lot of shared code and code duplication between the existing ezdm_summary_statistics function and its helpers and the newly proposed functions. Can you refactor to reduce the duplication as much as you can?

GidonFrischkorn:
Sure. I will have a look into this tomorrow afternoon.

GidonFrischkorn:
> There is a lot of shared code and code duplication between the existing ezdm_summary_statistics function and its helpers and the newly proposed functions. Can you refactor to reduce the duplication as much as you can?

Was the shared code primarily focussed on validation and the lapply structure for the grouped `.by` calculations. If yes, I have re-factored this by adding helpers that work for both functions. and reduce the place we have to maintain them in a single location.

venpopov:
A question about the flagging. Do we really need the probabilistic binomial flag generation? I don't find that useful generally and definitely not as a built-in option. It over-complicates the code for little value in my opinion

GidonFrischkorn:
> A question about the flagging. Do we really need the probabilistic binomial flag generation? I don't find that useful generally and definitely not as a built-in option. It over-complicates the code for little value in my opinion

Based on the p_contaminant curve flagging contaminants by threshold, say p_contaminant is larger than 0.5 basically results in adaptive thresholds for each subject for minimal and maximal RTs included in the analysis. When I worked on this I looked into the [Ratcliff & Tuerlincx (2002)](https://link.springer.com/article/10.3758/BF03196302) paper that discusses these outliers. And they argue that for slow RTs it is difficult to validate which RTs are proper outliers and which still belong to the accumulation process due to the long tail of the RT distribution. Thus, I thought instead of imposing hard bounds, parameter esimtation is likely less distorted with a probabilistic classification allowing for some slow RTs to not be flagged as RTs and avoid a bias in estimation of drift rate.

I see that this might be a very nuanced detail with respect to outlier treatment. And if you think the programming overhead is too big, we can also delete it. But from a theoeretical perspective it would be a superior treatment over hard bounds that flag all RTs below or above the bounds as contaminants.

venpopov:
ok, let me think about it

GidonFrischkorn:
I see your point that it can be outsourced if we already have p_contanminant for every observations. 

We could even consider just documenting it without providing an additional function.

venpopov:
that's what I'm thinking. I think the overall functions are a great addition to the package since dealing with outliers is an important part of measurement modeling. But I also want to make sure that the package stays focused and that we don't make our own future work with it more difficult by expanding the code a lot. The topic of outlier treatment is so broad and there is such a rich literature and range of approaches, that one could make a whole package dedicated just for that :)))

I like the option to document the use, and I haven't yet looked at your vignette, which seems excellent when I skimmed it, so perhaps that would be a good place to describe it.

GidonFrischkorn:
I fully agree!

Then it is better to focus on the mixture modelling and classifying the likelihood for being an outlier based on it.

I we want to maintain the outsourcing consequent we could also describe the validation logic for fast outliers.

This would likely also simplify the data helpers.

venpopov:
Glad we agree! Can you implement then and I’ll do a final pass after that tomorrow

GidonFrischkorn:
Sure. I will also take care of the failing checks.

These are, because I changed the warning for negative RTs to an error. And the tests also need a-station accordingly.

GidonFrischkorn:
I have now simplified the `flag_contaminant_rts` function and think it was a good decision to keep it that way. The vignette is shortened and the code is also reduced considerably.

I have also tried to remove unnecessary and self-explanatory comments.

venpopov:
That looks a lot cleaner! Will do final pass tomorrow

venpopov:
I read through the vignette to understand the whole picture - it's a very useful article and it will be a great addition to the website. Several things occur to me (separate comments):

### No need for separate contaminant_prop_upper contaminant_prop_lower in the flagging function

In the ezdm function that makes sense because you only get summary stats with wide format. For the flagging function you can return a single column even for the 4par version

```
head(flagged_4par)
     rt response strength correct obs_id contam_prob_upper contam_prob_lower
1 0.801    upper        8    TRUE      1        0.02569796                NA
2 0.680    upper        7    TRUE      2        0.01570458                NA
3 0.694    lower       19    TRUE      3                NA        0.02035604
4 0.582    lower       21   FALSE      4                NA        0.01445372
5 0.925    upper       19   FALSE      5        0.04240971                NA
6 0.605    upper       10    TRUE      6        0.01191150                NA
```

instead of two columns with missing values, just one is enough.

venpopov:
### No need for a 4par version for flagging 

As far as I can tell the 4par version is equivalent to using the .by = response argument in the 3par version. Unless I'm missing something, I suggest to remove the versioning from the flagging function entirely - no need for all the extra logic in the code if it already works with the generic .by split:

```r
flagged_4par <- flag_contaminant_rts(
  data = rr98_subset,
  rt = "rt",
  response = "response",
  version = "4par"
)

flagged_3par_resp <- flag_contaminant_rts(
  data = rr98_subset,
  rt = "rt",
  version = "3par",
  .by = "response"
)


head(flagged_4par)
head(arrange(flagged_3par_resp, obs_id))
```

```
     rt response strength correct obs_id contam_prob_upper contam_prob_lower
1 0.801    upper        8    TRUE      1        0.02569796                NA
2 0.680    upper        7    TRUE      2        0.01570458                NA
3 0.694    lower       19    TRUE      3                NA        0.02035604
4 0.582    lower       21   FALSE      4                NA        0.01445372
5 0.925    upper       19   FALSE      5        0.04240971                NA
6 0.605    upper       10    TRUE      6        0.01191150                NA

     rt response strength correct obs_id contam_prob
1 0.801    upper        8    TRUE      1  0.02569796
2 0.680    upper        7    TRUE      2  0.01570458
3 0.694    lower       19    TRUE      3  0.02035604
4 0.582    lower       21   FALSE      4  0.01445372
5 0.925    upper       19   FALSE      5  0.04240971
6 0.605    upper       10    TRUE      6  0.01191150
```

```
p4 <- ifelse(!is.na(flagged_4par$contam_prob_upper), flagged_4par$contam_prob_upper, flagged_4par$contam_prob_lower)
p3 <- arrange(flagged_3par_resp, obs_id)$contam_prob

all(p4 == p3)

# TRUE
```

venpopov:
## Vignette issues - wrong or incomplete code

The vignette has calls to ezdm without specifying variable:

```r
# 3. Fit EZDM model
fit <- bmm(
  formula = bmf(drift ~ condition, bs ~ 1, ndt ~ 1),
  data = ezdm_data,
  model = ezdm()
)
```

since ezdm requires variable names that call will fail, even if it is meant just as example.

---

The formula in that and other examples also uses incorrect parameter names - `bs` instead of `bound`

venpopov:
## Vignette issues - non-existing articles in cross-reference

```
### Cross-References

**Model-specific vignettes:**
- [EZ Diffusion Model](bmm_ezdm.html) - Full EZDM workflow and examples
- [Drift Diffusion Model](bmm_ddm.html) - DDM specification and fitting
- [Circular Shifted Wald](bmm_cswald.html) - csWald for circular responses

**Related topics:**
- [Getting Started with bmm](bmm.html) - Package overview
- [Formula Specification](bmm_formula.html) - Building bmmformula objects
- [Model Comparison](bmm_model_comparison.html) - Comparing model fits
```

I suspect you might be pre-emptively using linsk for upcoming models - still, those should be added when they exist. But in the related topics, bmm.html doesn't exist and neither does bmm_model_comparison.html

---

For the actual references, they should be in REFERENCES.bib and inserted with bibkeys, and no manual ref section at the end, like we do it for the other vignettes

venpopov:
## .by argument shouldn't change the data order

While checking if the 4par flagging is identical to the 3par with .by=response I noticed that using .by lead to rearrangement of the data rows. This shouldn't happen. 

I suggest to simplify this function and:
-  take a single vector of values (RTs)
- not deal with .by or response variables
- output the contam_prob vector

Then it can be used as any other data processing function, either with base R or tidyverse by the user like:

```r
mydata |>
   mutate(contam_prob = flag_contaminant_rts(resp_time))
```
and they can group however they like:

```r
mydata |>
   group_by(subject, response) |>
   mutate(contam_prob = flag_contaminant_rts(resp_time))
```

You can also attach the diagnostics as an attribute of the returned vector, instead of having two different return options.

that way we stick to "single responsibility" - the function does one thing and does it well. The data wrangling and splitting is handled separately like any other part of a data processing pipeline

GidonFrischkorn:
While thinking about your comments. I considered if we should take a similar approach to the `ezdm_summary_stats function.

We could implement it similarly, providing robust summary stats for the vector of `rt` and `responses` for a single subject in one experimental condition. For data pre-processing users can then use `group_by` combined with `reframe` to obtain the summary stats for the grouping variables they need:

```
# Per-subject summary stats for 3par ezdm
summary_data <- mydata |>
  group_by(subject) |>
  reframe(ezdm_summary_stats(rt, response))

# Fit the model
model <- ezdm(mean_rt = "mean_rt", var_rt = "var_rt",
              n_upper = "n_upper", n_trials = "n_trials")
fit <- bmm(bmf(drift ~ 1, bound ~ 1, ndt ~ 1),
           data = summary_data, model = model)

```

What do you think, should I include this re-factoring of ezdm_summary_stats() in this pull request?

venpopov:
Agreed, that would be my choice too.

venpopov:
This looks great now! Very clean and simple, and I also like the shorter and easier to read vignette. I think it's ready, except for one question:

- What should the default bounds on the uniform component be? 

They are currently [0.1, 3]. Is that based on a recommendation from the papers you referenced? Naturally changing the upper bound changes the probability that an observation is considered a contaminant.

You also allow ("min", "max"), which seems to be a better default, although I can see arguments for both. 

How does the current mixture approach deal with observations that are above the max bound of the uniform? When I plot the probability that something is flagged as a function of RT on the vignette example data, with an upper bound of 2, I get

```r
x <- flag_contaminant_rts(dat$rt, contaminant_bound = c(0.1, 2))
plot(dat$rt, x)
abline(v = 2)
```
Screenshot 2026-02-26 at 11 57 07

If the model is strictly using a mixture with a uniform bounded on [0.1, 2], as stated in the vignette, $f_{uniform}$ will be 0 outside of that range. But the probability clearly doesn't drop to 0 after 2, so it must be doing something differrent (or I am misunderstanding what it does).

---
In terms of the code itself, this is good to go. I just want to clarify the default bounds and the mixture approach before we merge it

GidonFrischkorn:
For computational efficiency the uniform density was calculated as `1/(b-a)` and applied to all RTs irrespective whether they are inside or outside the bounds. I adapted this. Thanks for catching this.

With respect to the default choice, I have a tendency to go for the data-driven bounds to avoid critical RTs not being flagged as outliers when they are outside the bounds and necessarily captured by the RT-like distribution.

The other option would be to classify any RT outside the bounds as contaminant. This would basically implement filtering the RTs to only RTs inside the bounds prior to applying the mixture-fitting. But this is something that users can and should do prior to using these summary functions.

So, my tendency is to default to the data-driven bounds and add a note in the vignette that users should filter unplausible RTs (< 0.1 s or larger > 10s) prior to applying these functions.

What do you think?

venpopov:
The data-driven bounds also make most sense to me. It is a tricky question because the bound determines the relative weight on the uniform likelihood - the greater the bound is, the less likely an observation is to come from the uniform component. But that's an issue for any bound and any mixture approach. So data-driven makes sense, with the option to specify custom bounds as now. It would be useful to have a warning if the max bound for the uniform component is less that the data max so that the user knows that any observations greater than the bound will be classified as not outliers

GidonFrischkorn:
Now, everything should be ready. I also ensured robustness of some tests.

GidonFrischkorn:
image

venpopov (approved):

venpopov:
Awesome, this is a great addition to the package. Will merge now.

PR #314, opened 2026-02-18
Add logo.png

GidonFrischkorn:
#### Summary
this is just sugar for the package. I created a logo and shared it with you a while ago. Here I have added it via `usethis::use_logo()`

#### Tests


venpopov (approved):
looks great

PR #316, opened 2026-02-19
Update AGENTS.md with new test and development folders

GidonFrischkorn:
#### Summary
- I have set up a symlink solution to synchronize my development files between the two laptops I am working with (UZH/Luzern) and wanted to reflect this in the AGENTS.md file.
- I have also added your local .gitignore solution to the file
- I also wanted to talk to you about sharing some of the development planning with you, but I thought this does not need to be synchronized over GitHub

#### Tests

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -119,6 +119,8 @@ Stan code snippets go in `inst/stan_chunks/*.stan` - these are injected into brm
 Test organization:
 - `tests/testthat/` - automated unit tests (run by CI)
 - `tests/internal/` - long-running manual tests
+- `bmm-dev/feature_tests/` - exploratory/development scripts (synced via SWITCHdrive, not tracked by git)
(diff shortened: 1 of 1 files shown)


venpopov (changes requested):
I see why this is useful for you, but the problem is that this file structure is specific to your local files and its not true for the repo in general or my local copy. It will likely confuse my agent.

I suggest that you:

add another file, something like LOCAL-AGENTS.md
add this file to .gitignore
add one line mention in AGENTS.md that the agent must read LOCAL-AGENTS.md if it exists.
That way I can also specify my own local instructions, and the only thing we share is the general file which includes the instruction to read the local file.

Can you try it and see if that work for your setup?

GidonFrischkorn:
This works fine for me. I have adapted the changes accordingly.

venpopov (approved):

PR #348, opened 2026-03-31
Fix swald_lccdf returning -inf instead of 0 for t <= ndt

GidonFrischkorn:
While working on some tutorial scripts for the evidence accumulation models I stumbled across a bug in the `swald_lccdf` function. In swald_lccdf (cswald_helper_functions.stan:29)the log-survival function returned negative_infinity() when rt - ndt <= 0. The correct return value is 0 (log of survival probability 1), since the decision process hasn't started yet.

The density function `swald_lpdf` correctly returns -infinity() for this case (density = 0), but the survivor function must return 0 (survival = 1).


GidonFrischkorn:
Together with fixing the citation in the description we can potentially submit this to CRAN as `1.3.1`

venpopov (approved):

venpopov:
@GidonFrischkorn can you add a news item for the fix to the release PR? #351

PR #359, opened 2026-06-14
Add `sdt_binary` to bmm

GidonFrischkorn:
## Summary

Adds `sdt_binary()` — a binary Signal Detection Theory model for yes/no detection and recognition tasks with aggregated hit / false-alarm counts. This is the first in a planned series of response-format-specific SDT constructors (rating, m-AFC, ranking to follow in separate PRs), built self-contained off `develop`.

## What's included

- **Model**: `sdt_binary(response, stimulus, n_trials, dist, ...)` with a binomial likelihood, estimating `dprime` and `criterion`, optional unequal-variance scaling via `sdratio`, and selectable noise distribution (`normal`, `logistic`, `gumbel_min`, `gumbel_max`).
- **Distribution functions**: `dsdt_binary()`, `rsdt_binary()`.
- **Utilities**: `sdt_dprime()`, `sdt_criterion()` (sensitivity / bias from hit & FA rates).
- **Shared SDT numerics** (`.SDT_DISTS` registry, `.sdt_eta`, `.sdt_cdf`) in `distributions.R`; pipeline helpers in `helpers-sdt.R`.
- **Stan**: `sdt_binary_funs.stan` with numerically stable log-CDF dispatch.
- **Tests**: full coverage of construction, validation, d/r functions, and stancode generation; the package-wide `check_data()` test extended to cover `sdt_binary`.

## Deferred to follow-up PRs

ROC/AUC analysis (`roc_sdt`/`auc_sdt`) and plotting are irreducibly multi-model (they need rating-format data), so they land with the rating PR rather than here.

## Validation

- Parameter recovery verified with an independent (non-circular) simulator.


GidonFrischkorn:
## Added dataset: `broeder_schuetz_2009_e3` (unequal-variance binary SDT)

### Why include it
bmm ships datasets that showcase each model's key functionality, and `sdt_binary`'s distinctive feature is the optional **unequal-variance** parameter `sdratio` (the signal/noise SD ratio, i.e. the zROC slope). None of the existing datasets can demonstrate it: `data_color_judgement_task` manipulates **difficulty (sensitivity / d′)**, not the decision criterion — and `sdratio` is only identifiable when the **criterion varies across conditions** while d′ and the variance ratio stay fixed. A single (HR, FAR) point cannot separate a wider signal distribution from a larger d′.

`broeder_schuetz_2009_e3` (Bröder & Schütz, 2009, Exp. 3) fills that gap: 40 subjects, binary old/new recognition under a **5-level base-rate manipulation** that sweeps the criterion across 5 conditions and traces a 5-point binary ROC per subject — exactly the structure `sdratio` needs.

- **Format:** tidy aggregated counts (400 rows = 40 subjects × 5 conditions × 2 stimulus types; `id, condition, stimulus, n_old, n_trials`), following the `oberauer_lewandowsky_2019_e1` aggregated-count precedent.
- **Provenance:** counts digitised from the published frequencies; `@source` cites the article in `R/data.R`. Third-party data, included with attribution as with `zhang_luck_2008`.

### Estimating `sdratio` — specification
The identifying spec lets the criterion vary while holding d′ and the variance ratio constant across conditions:
```r
bmm(
  bmf(
    dprime    ~ 1 + (1 | id),               # constant across conditions
    criterion ~ 0 + condition + (1 | id),   # criterion varies by base rate
    sdratio   ~ 1                            # estimated (overrides the fixed default)
  ),
  data  = broeder_schuetz_2009_e3,
  model = sdt_binary("n_old", "stimulus", "n_trials")
)
```
On these data this yields a signal/noise SD ratio of **1.46 [1.25, 1.72]** (zROC slope ≈ 0.69) — the classic recognition-memory pattern (signal distribution wider than noise). In simulation with a known ratio of 1.3, the parameter recovers cleanly (log 0.262 → 0.241 [0.197, 0.284]).

### Hierarchical `sdratio` — possible, but weakly identified here
Because each subject has all 5 criterion conditions, a per-subject zROC slope is identified in principle, and `sdratio ~ 1 + (1 | id)` **does not break identification**:
- converges cleanly (0 divergences, R̂ ≤ 1.02);
- the population estimate is essentially unchanged (0.39 [0.23, 0.55] on the log scale);
- the between-subject SD is estimable: **0.137 [0.015, 0.272]** (CI excludes 0).

But it is only **weakly** identified at the subject level: per-subject `sdratio` 95% CIs are ~0.59 wide versus a between-subject estimate spread of ~0.30, so the hierarchical layer mostly shrinks toward the population value. (The variance ratio is the hardest SDT parameter to pin down, and several cells have few trials.)

GidonFrischkorn:
Here are some plots on a small parameter recovery simulation across all implemented noise distributions and variable number of trials. in each design cell I ran 100 repetitions that randomly sampled the hyper parameter means from a fixed range.

Hyperparameter Means are unbiased across all numbers of trials with a small tendency for overestimation on d-prime, but the 95% CIs all include zero, so this seems negligible from my perspective

hyperpars_bias_SDT_Binary

The concordance correlations coefficient (combining recovery of relative position, scale bias, and location bias) is excellent for hyperparmaeter recovery.

hyperpars_ccc_SDT_Binary

And here are plots of true subject parameters against estimated subject parameters from a subset of simulations across the different conditions. For subject parameter recover increasing number of trials provides a large benefit in recovery, for both the criterion:

scatter_SDT_Binary_criterion

same for dprime:

scatter_SDT_Binary_dprime

If I should share the full results with you, just say so. Then I can send you a zip-Archive.

GidonFrischkorn:
And some added info. Some of the infrastructure such as the distrbution functions for different noise distributions are outsourced to a general helper file for all `sdt` models.

If you have a different proposition how to handle these helpers, feel free to suggest another solution. I just thought these functions should also not live in one of the specific model files, if these are used by more than on of the `sdt_*` functions for different response formats.

PR #360, opened 2026-06-15
add `sdt binary`

### LINKED ISSUE #108: implement SDT models
Because of the Master thesis I am supervising, I had a look into options for implementing SDT models and found out that brms already supports basically all link functions that are relevant for this: mostly `probit` for SDT with normal noise and `cloglog` for SDT with Gumbel noise. Additional options would be `cauchit` or `logit` for SDT models with Cauchy or Logistic noise. But I don't think these are really relevant. These links are implement for both `bernoulli` or `binomial` distributions for single threshold SDT models, as well as for the `cumulative` distribution needed for the multiple threshold/confidence SDT models.

I was thinking about implementing two SDT models for 1) the single threshold case and 2) the multiple threshold case and in addition to the response variable let the users specify which noise distribution they want to assume (i.e. "normal" or "gumbel"). For the multiple threshold models, there would be an additional option for unequal variance models, providing the option to predict a "discrimination" parameter.

Would you mind, if I take a shot at these implementations? Or do you want to do this? Just because originally, we discussed that you would take the lead in this.

Apart from that I think these models should support all of the `brms` functionality, given that the distributions and link functions are natively supported by `brms`.

GidonFrischkorn:
## Summary

Adds `sdt_binary()` — a binary Signal Detection Theory model for yes/no detection and recognition tasks with aggregated hit / false-alarm counts. This is the first in a planned series of response-format-specific SDT constructors (rating, m-AFC, ranking to follow in separate PRs), built self-contained off `develop`.

## What's included

- **Model**: `sdt_binary(response, stimulus, n_trials, dist, ...)` with a binomial likelihood, estimating `dprime` and `criterion`, optional unequal-variance scaling via `sdratio`, and selectable noise distribution (`normal`, `logistic`, `gumbel_min`, `gumbel_max`).
- **Distribution functions**: `dsdt_binary()`, `rsdt_binary()`.
- **Utilities**: `sdt_dprime()`, `sdt_criterion()` (sensitivity / bias from hit & FA rates).
- **Shared SDT numerics** (`.SDT_DISTS` registry, `.sdt_eta`, `.sdt_cdf`) in `distributions.R`; pipeline helpers in `helpers-sdt.R`.
- **Stan**: `sdt_binary_funs.stan` with numerically stable log-CDF dispatch.
- **Tests**: full coverage of construction, validation, d/r functions, and stancode generation; the package-wide `check_data()` test extended to cover `sdt_binary`.

## Deferred to follow-up PRs

ROC/AUC analysis (`roc_sdt`/`auc_sdt`) and plotting are irreducibly multi-model (they need rating-format data), so they land with the rating PR rather than here.

## Validation

- Parameter recovery verified with an independent (non-circular) simulator.


GidonFrischkorn:
## Added dataset: `broeder_schuetz_2009_e3` (unequal-variance binary SDT)

### Why include it
bmm ships datasets that showcase each model's key functionality, and `sdt_binary`'s distinctive feature is the optional **unequal-variance** parameter `sdratio` (the signal/noise SD ratio, i.e. the zROC slope). None of the existing datasets can demonstrate it: `data_color_judgement_task` manipulates **difficulty (sensitivity / d′)**, not the decision criterion — and `sdratio` is only identifiable when the **criterion varies across conditions** while d′ and the variance ratio stay fixed. A single (HR, FAR) point cannot separate a wider signal distribution from a larger d′.

`broeder_schuetz_2009_e3` (Bröder & Schütz, 2009, Exp. 3) fills that gap: 40 subjects, binary old/new recognition under a **5-level base-rate manipulation** that sweeps the criterion across 5 conditions and traces a 5-point binary ROC per subject — exactly the structure `sdratio` needs.

- **Format:** tidy aggregated counts (400 rows = 40 subjects × 5 conditions × 2 stimulus types; `id, condition, stimulus, n_old, n_trials`), following the `oberauer_lewandowsky_2019_e1` aggregated-count precedent.
- **Provenance:** counts digitised from the published frequencies; `@source` cites the article in `R/data.R`. Third-party data, included with attribution as with `zhang_luck_2008`.

### Estimating `sdratio` — specification
The identifying spec lets the criterion vary while holding d′ and the variance ratio constant across conditions:
```r
bmm(
  bmf(
    dprime    ~ 1 + (1 | id),               # constant across conditions
    criterion ~ 0 + condition + (1 | id),   # criterion varies by base rate
    sdratio   ~ 1                            # estimated (overrides the fixed default)
  ),
  data  = broeder_schuetz_2009_e3,
  model = sdt_binary("n_old", "stimulus", "n_trials")
)
```
On these data this yields a signal/noise SD ratio of **1.46 [1.25, 1.72]** (zROC slope ≈ 0.69) — the classic recognition-memory pattern (signal distribution wider than noise). In simulation with a known ratio of 1.3, the parameter recovers cleanly (log 0.262 → 0.241 [0.197, 0.284]).

### Hierarchical `sdratio` — possible, but weakly identified here
Because each subject has all 5 criterion conditions, a per-subject zROC slope is identified in principle, and `sdratio ~ 1 + (1 | id)` **does not break identification**:
- converges cleanly (0 divergences, R̂ ≤ 1.02);
- the population estimate is essentially unchanged (0.39 [0.23, 0.55] on the log scale);
- the between-subject SD is estimable: **0.137 [0.015, 0.272]** (CI excludes 0).

But it is only **weakly** identified at the subject level: per-subject `sdratio` 95% CIs are ~0.59 wide versus a between-subject estimate spread of ~0.30, so the hierarchical layer mostly shrinks toward the population value. (The variance ratio is the hardest SDT parameter to pin down, and several cells have few trials.)

GidonFrischkorn:
Here are some plots on a small parameter recovery simulation across all implemented noise distributions and variable number of trials. in each design cell I ran 100 repetitions that randomly sampled the hyper parameter means from a fixed range.

Hyperparameter Means are unbiased across all numbers of trials with a small tendency for overestimation on d-prime, but the 95% CIs all include zero, so this seems negligible from my perspective

hyperpars_bias_SDT_Binary

The concordance correlations coefficient (combining recovery of relative position, scale bias, and location bias) is excellent for hyperparmaeter recovery.

hyperpars_ccc_SDT_Binary

And here are plots of true subject parameters against estimated subject parameters from a subset of simulations across the different conditions. For subject parameter recover increasing number of trials provides a large benefit in recovery, for both the criterion:

scatter_SDT_Binary_criterion

same for dprime:

scatter_SDT_Binary_dprime

If I should share the full results with you, just say so. Then I can send you a zip-Archive.

R/distributions.R
```diff
@@ -1945,3 +1945,202 @@ neg_loglik <- function(x, params, distribution, weights = NULL) {
...
+#' @rdname sdt_binary_dist
+#' @export
+#' @examples
+#' # Generate binary SDT data for a design
+#' dat <- expand.grid(id = 1:20, stimulus = c(0L, 1L))
+#' dat$n_trials <- 100L
+#' dat$n_old <- rsdt_binary(nrow(dat), dat$n_trials, dat$stimulus,
+#'                          dprime = 1.5, criterion = 0.2)
+#' head(dat)
+rsdt_binary <- function(n, n_trials, stimulus, dprime, criterion,
```
venpopov: What is the role of the stimulus argument? I don't see a purpose for it for the random generation function and was confused when I tried to use it. I

R/model_sdt_binary.R
```diff
@@ -0,0 +1,225 @@
...
+      name = "Signal Detection Theory (Binary)",
+      citation = glue(
+        "Green, D. M., & Swets, J. A. (1966). Signal detection theory ",
+        "and psychophysics. Wiley."
+      ),
+      version = "NA",
+      requirements = requirements,
+      parameters = parameters,
+      links = list(dprime = "identity", criterion = "identity",
+                   sdratio = "identity"),
```
venpopov: in the documentation you say that sdratio has a log link

R/distributions.R
venpopov: Oh, I get it after seeing the example. It would help to have it more explicitly described in the documentation for the argument

R/model_sdt_binary.R
```diff
@@ -0,0 +1,233 @@
...
+                              links = NULL, call = NULL, ...) {
+  parameters <- list(
+    dprime = "Sensitivity: distance between signal and noise distributions",
+    criterion = "Response bias: location of decision boundary",
+    sdratio = paste0(
+      "SD ratio: the signal-to-noise standard deviation ratio, estimated on ",
+      "the log scale, so 1 means equal variance"
+    )
+  )
+  default_priors <- list(
```
venpopov: There is an error when trying to use the default_prior function:

``` r
library(bmm)

dat <- expand.grid(id = 1:20, stimulus = c(0L, 1L))
dat$n_trials <- 100L
dat$n_old <- rsdt_binary(nrow(dat), dat$n_trials, dat$stimulus,
                         dprime = 1.5, criterion = 0)

model <- sdt_binary(
  response = "n_old",
  stimulus = "stimulus",
  n_trials = "n_trials",
  dist = "normal"
)

fit <- bmm(
  formula = bmf(dprime ~ 1, criterion ~ 1),
  data = dat,
  model = model,
  cores = 4,
  backend = "cmdstanr",
  refresh = 0
)
#> Start sampling
#> Running MCMC with 4 parallel chains...
#> 
#> Chain 1 finished in 0.1 seconds.
#> Chain 2 finished in 0.1 seconds.
#> Chain 3 finished in 0.1 seconds.
#> Chain 4 finished in 0.1 seconds.
#> 
#> All 4 chains finished successfully.
#> Mean chain execution time: 0.1 seconds.
#> Total execution time: 0.2 seconds.
#> Loading required namespace: rstan

# this doesn't work
default_prior(fit$formula, data = dat, model = model)
#> Error in `validate_data()`:
#> ! The following variables can neither be found in 'data' nor in 'data2':
#> 'dist_type'

# this works
brms::prior_summary(fit)
#>           prior     class coef group resp      dpar nlpar lb ub tag source
#>  normal(0, 1.5) Intercept                 criterion                   user
#>    normal(1, 1) Intercept                    dprime                   user
#>     constant(0) Intercept                   sdratio                   user
```

<sup>Created on 2026-08-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/model_sdt_binary.R
```diff
@@ -0,0 +1,233 @@
...
+#'
+#' fit <- bmm(
+#'   formula = bmf(dprime ~ 1, criterion ~ 1),
+#'   data = dat,
+#'   model = model,
+#'   cores = 4,
+#'   backend = "cmdstanr"
+#' )
+#'
+#' # Unequal-variance binary SDT
```
venpopov: we should have some warnings about identifiability of sdratio. In the general setup where you have a simple yes/no sdratio is not identifiable, unless criterion varies among participants but not dprime (both varying is also problematic), or if you have manipulations of response bias

PR #361, opened 2026-06-15
add `sdt_mafc`

### LINKED ISSUE #108: implement SDT models

GidonFrischkorn:
Adds `sdt_mafc()` — a Signal Detection Theory model for m-Alternative Forced Choice tasks, stacked on the `new_model/sdt_binary` branch. This is the second in the planned series of response-format-specific SDT constructors. It extends the binary foundation to accuracy data (correct-response counts) with support for all four noise distributions and any number of alternatives ≥ 2.

## What's included

- **Model**: `sdt_mafc(response, n_trials, m, dist, ...)` with a binomial likelihood on correct-response counts, estimating `dprime` only — m-AFC eliminates response bias so no criterion is needed.
- **P(correct) computation** per noise distribution:
  - `gumbel_min`: closed-form softmax — `1 / (1 + (m−1)·exp(−d′))`
  - `gumbel_max`: closed-form Gamma ratio — `Γ(1+e^{−d′})·Γ(m) / Γ(m+e^{−d′})`
  - `normal`: Gauss-Hermite quadrature; closed form `Φ(d′/√2)` at m = 2
  - `logistic`: GL-64 Gauss-Legendre quadrature on [0, 1] via `sdt_quantile`
- **Distribution functions**: `dsdt_mafc()`, `rsdt_mafc()` with recycling-based vectorisation (longest-argument recycling, no `n_subjects` argument).
- **Shared Stan chunk** (`sdt_dist_funs.stan`): `sdt_quantile()` dispatch added alongside the CDF/log-CDF functions from the binary PR. Both `sdt_binary` and `sdt_mafc` load this shared chunk before their own lpmf, so the four-distribution numerics are defined once.
- **Stan**: `sdt_mafc_funs.stan` with `mafc_pc(dprime, m, dist_type)` and `sdt_mafc_lpmf`.
- **Tests**: constructor, `check_data`, d/r recycling, per-distribution oracle checks (independent integrals/binomial sums), stancode assembly, and mock pipeline for all four distributions (142 checks).

## Deferred to follow-up PRs

No criterion, no threshold structure — rating-format SDT lands in its own PR. ROC/AUC analysis remains deferred as it needs rating data.

## Validation

- Parameter recovery: d′ recovers across all 48 conditions (4 nTrials × 3 m × 4 dist), 100 replications each, 50 subjects. Median CCC = 0.994, RMSE = 0.063, 95% CI coverage = 95%. Data generated via independent inverse-CDF argmax sampling — no shared code with the fitted likelihood, so generation and fitting bugs cannot cancel.


GidonFrischkorn:
Here are some of the recovery results:

1. CCC for hyper parameter means across all noise distributions and different levels of m-Alternatives:

hyperpars_ccc_SDT_mAFC

2. Bias of hyper parameter means across noise distributions and different levels of m-Alternatives:

hyperpars_bias_SDT_mAFC

3. True vs Recovered subject-level d` for m = 4 across different noise distributions and numbers of trials of a subset of the subject parameters across the 100 replications in each design cell
 
scatter_SDT_mAFC_dprime_4

PR #374, opened 2026-07-04
Improve SDM denominator evaluation and threading

chenyu-psy:
This PR makes two exact computational changes to the SDM implementation.

1. The SDM normalization constant is now evaluated once per contiguous run of identical SDM predictors, instead of checking `c[n]` and `kappa[n]` equality inside Stan for every trial. This removes the Stan-side real-valued equality check and avoids repeated denominator calculations in sorted designs.

2. SDM now supports optional within-chain threading through the existing `threads` argument. Both `threads = 2` and `threads = brms::threading(2)` are supported. Default non-threaded behavior is unchanged.

In a local 30-subject × 3-condition short-chain benchmark, the original implementation took 87.6s. The run-level denominator version took 77.3s (11.7% faster). With threading enabled, the same benchmark took 47.2s with 2 threads per chain (46.1% faster than the original implementation) and 45.9s with 3 threads per chain (47.6% faster).

In a local parameter-recovery benchmark using 30 subjects, 100 trials per subject, 3 conditions, 4 chains, 1000 warmup iterations, and 3000 total iterations per chain, the original implementation took 520.3s. The run-level denominator version took 499.7s (4.0% faster). The threaded version with 8 total CPU threads took 336.1s (35.4% faster than the original implementation). Divergences were 0, and fixed-effect recovery coverage was 6/6 in the recovery runs.


R/model_sdm.R
```diff
@@ -186,3 +202,42 @@ posterior_predict_sdm_simple <- function(i, prep, ...) {
...
+
+sdm_predictor_vars <- function(data, formula) {
+  dpars <- names(formula)
+  predictors <- rhs_vars(formula)
+  predictors <- predictors[not_in(predictors, dpars)]
+  predictors[predictors %in% colnames(data)]
+}
+
+sdm_stanvar_int_array <- function(x, name, size) {
+  out <- brms::stanvar(x = as.integer(x), name = name)
```
GidonFrischkorn: For a model without predictors (`c ~ 1, kappa ~ 1`, the documented `?sdm` example), `sdm_run_metadata()` returns `G_sdm_runs = 1`, so `sdm_run_start` and `sdm_run_count` are length-1 integer vectors. Stan declares them `array[G_sdm_runs] int`, but cmdstanr serializes a plain length-1 integer as a scalar rather than an array.

The result is a data-dimension mismatch, and the model does not run. The new tests only cover the two-condition case (`G = 2`), where the arrays serialize fine, so this path is currently untested.

It would also be worth adding a test that builds standata for an intercept-only SDM (`c ~ 1, kappa ~ 1`) and asserts `G_sdm_runs == 1` with `sdm_run_start` and `sdm_run_count` present as arrays, so this case stays covered.

R/model_sdm.R
```diff
@@ -186,3 +202,42 @@ posterior_predict_sdm_simple <- function(i, prep, ...) {
...
+  run_start <- c(1L, which(run_id[-1] != run_id[-length(run_id)]) + 1L)
+  run_count <- diff(c(run_start, nrow(data) + 1L))
+  list(
+    G_sdm_runs = length(run_start),
+    sdm_run_start = as.integer(run_start),
+    sdm_run_count = as.integer(run_count)
+  )
+}
+
+sdm_predictor_vars <- function(data, formula) {
```
GidonFrischkorn: The block `rhs_vars(formula)` → drop dpars → intersect `colnames(data)` now lives in three places:

- `order_data_query()` — `R/utils.R:310-312`
- `is_data_ordered()` — `R/helpers-data.R:282-284`
- `sdm_predictor_vars()` — `R/model_sdm.R:228-230` (new)

The aggregation here is currently only correct because `sdm_predictor_vars()` returns exactly the predictor set the data was sorted by. If one of these copies is later changed and the others are not, run detection would diverge from the sort order and apply the wrong denominator to a block of rows. That is a parameter-dependent likelihood error with no error message, which is the kind of bug that is hard to catch later. I would pull this into a single helper and call it from all three sites, so the guarantee cannot drift. What do you think @venpopov

R/model_sdm.R
```diff
@@ -137,9 +138,24 @@ configure_model.sdm <- function(model, data, formula) {
...
+    "sdm_simple_likelihood_threaded.stan"
+  } else {
+    "sdm_simple_likelihood.stan"
+  }
+  stan_likelihood <- read_lines2(paste0(sc_path, "/", likelihood_file))
+  stan_tdata_pll_args <- if (sdm_use_threaded_likelihood()) {
+    "data matrix COSN"
+  }
+  run_metadata <- attr(data, "sdm_run_metadata")
+  if (is.null(run_metadata)) {
```
GidonFrischkorn, line 150: this never fires in the normal pipeline, since `check_data.sdm()` always sets the attribute.

Feel free to drop it, or leave a one-line comment naming the direct-call path it is meant to guard.

GidonFrischkorn (changes requested):
This looks great overall. The run-level denominator aggregation and the threading
both hold up.

A few changes are needed before this can merge, one I would strongly recommend, and two that are optional, but would improve long term maintainability.

Given that @venpopov has done most of the work around the `sdm`, I would also wait for him to have a look over the changes and approve them

R/model_sdm.R
chenyu-psy: Fixed in the follow-up commit. `sdm_stanvar_int_array()` now preserves array shape even when `G_sdm_runs == 1`, so intercept-only SDM formulas no longer pass plain scalar-looking length-1 integers to Stan.

I also added standata regression coverage for `c ~ 1, kappa ~ 1`, checking that `G_sdm_runs == 1` and that `sdm_run_start` / `sdm_run_count` remain arrays.

R/model_sdm.R
chenyu-psy: Agreed. I pulled the predictor-selection logic into a shared internal helper and now use it from `order_data_query()`, `is_data_ordered()`, and `sdm_run_metadata()`. That keeps the sort-order predictor set and run-detection predictor set tied to the same code path.

venpopov:
Looks good on first pass. Thanks @chenyu-psy! I will review more carefully and let you know if any changes are needed. Btw, the existing comparison already avoided repeated computation of the normalization constant - that was the entire point of it. The first of your changes makes it cleaner and is a nice addition but it doesn't introduce "avoids repeated denominator calculations in sorted designs" as the code already did that. The threading support is very nice.

venpopov (commented):
Thanks @chenyu-psy — the core of this PR is solid and the benchmarks hold up. I verified the math and mechanics directly on the branch:

- The run-level denominator is exact, and the threaded slice attribution (each run's `run_count[g] * z` credited to the slice containing `run_start[g]`) sums exactly to the serial total across slices.
- The slice-local index mapping `n = run_start[g] - start + 1` is correct against the `partial_log_lik` code brms actually generates (slice-local `c`/`kappa`, global `start`/`end`), and the `-(log2()+log(pi()))*N` term uses slice-local `N`, which totals to the global N.
- Run detection is change-point based, so it stays correct on unsorted data (just with more runs). All 28 new tests pass locally.

Testing did surface three bugs, all sharing one root cause: the run metadata is computed R-side and baked into the Stan data as stanvar *values*, so it can drift out of sync with the data brms actually uses (the old Stan-side `c[n] != c[n-1]` check was immune to all three by construction). **No action needed on your side — we'll implement the fixes in a follow-up on top of this PR.** Documenting them here for the record:

### 1. `update()` reuses stale run metadata → silently wrong posteriors

`update.bmmfit()` recomputes `config_args` (with fresh stanvars) but never forwards them to brms; `brms::update.brmsfit()` then falls back to `object$stanvars`, i.e. the values baked in when the model was first fit. Verified end-to-end: fit an intercept-only sdm on N = 20, then `update(fit, newdata = <40 rows>)` — the refit uses `N = 40` but `sdm_run_count = 20`, so the normalizing constant enters at half its correct weight, with no error or warning. The roxygen example in `update.R` hits exactly this case. Formula-only updates that add predictors (e.g. `c ~ 1` → `c ~ 0 + condition`) are affected too, and recompilation doesn't help because brms revalidates the *old* stanvars.

Fix: pass `stanvars = new_fit_args$stanvars` in the `NextMethod("update", ...)` call in `update.bmmfit()`.

### 2. NA rows misalign the metadata → silently wrong likelihood

The metadata is computed in `check_data.sdm()` on the full data, but brms drops incomplete cases afterward. Reproduced: 20 rows with one NA response → `N = 19` but `sum(sdm_run_count) = 20`, with `run_start` referring to pre-drop row positions, so every run boundary after the dropped row is shifted. Depending on where the NAs fall this is either silently wrong or a Stan index error.

Fix: compute the metadata on complete cases of the model variables (mirroring brms's exclusion), plus the Stan-side check below as a backstop.

### 3. Numeric `options(brms.threads = 2)` generates a self-inconsistent model

brms's `validate_threads()` accepts a bare number for the global option, but `sdm_use_threaded_likelihood()` requires a list — so brms wraps the likelihood in `reduce_sum` while bmm embeds the *non-threaded* chunk inside `partial_log_lik`, which indexes slice-local `c`/`kappa` with global `run_start` positions. Verified: the generated code contains both `target += reduce_sum(partial_log_lik_lpmf, ...)` and `ptarget += sdm_simple_run_ldenom(...)`, and at runtime chains fail with index-out-of-range exceptions (scheduler-dependent). To be fair, any global `brms.threads` setting also broke sdm before this PR, so this is an inherited gap rather than a new regression — but with threading now officially supported it should be closed.

Fix: normalize numerics before the check, e.g. `if (is.numeric(threads)) threads <- brms::threading(threads)`, mirroring `validate_threads()`.

### Hardening + minor (also for the follow-up)

- A Stan `transformed data` consistency check (`sum(sdm_run_count) == N`, `run_start` strictly increasing and in bounds) so any future metadata drift fails loudly at data-load time instead of silently corrupting the posterior — this would have caught 1 and 2 outright.
- Regression tests: `sum(sdm_run_count) == sdata$N` on NA-containing data, and an `update(newdata)` test.
- A NEWS.md entry for the feature + performance change.

Again — none of this needs changes from you; we'll take it from here. Thanks for the careful implementation and the thorough benchmarks!

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

Issue #354, opened 2026-04-05
`create_initfun` parameter matching fails when model parameter names are substrings of each other


GidonFrischkorn:
## Description

`create_initfun` in `R/helpers-inits.R` (line 50) maps Stan parameter names to model parameter names using a substring grep:

```r
parameter <- model_pars[unlist(lapply(paste0("_", model_pars), grepl, x = spar))]
```

This fails when one model parameter name is a substring of another. The pattern `_s` (for parameter `s`) matches inside `Intercept_sim` or `Intercept_similar`, causing incorrect init ranges and crashes (`if (parameter == "Intercept")` errors when multiple matches return a length > 1 vector).

## Concrete failure cases

| Stan param | Pattern | Matches | Should match |
|------------|---------|---------|-------------|
| `Intercept_sim` | `_s` | `s` (wrong) | `sim` |
| `Intercept_similar` | `_s` | `s` (wrong) | `similar` |
| `Intercept_correct` | `_c` | `c` (wrong) | `correct` |
| `Intercept_abc` | `_a` | `a` (wrong) | `abc` |
| `Intercept_mu1` | `_mu` | `mu` (wrong) | `mu1` |

## Affected parameters

Single-letter model parameter names with high collision risk:
- `s` (cswald, ezdm, rdm, lnr) — collides with any category starting with "s"
- `c` (sdm, imm, m3) — collides with "correct", "corr", etc.
- `a` (imm, m3) — collides with "acc", "alt", etc.
- `A` (rdm) — collides with any category starting with "A"

Multi-letter prefix collisions:
- `mu` vs `mu1`, `mu2` (mixture models)
- `kappa` vs `kappa2` (imm/mixture)

## Discovered independently on two branches

- **Branch 352 (LNR):** `s` matched inside custom category `similar`. Fixed with perl lookahead `(?=_|$)` + first-match fallback.
- **Branch 353 (RDM):** `s` matched inside custom category `sim`. Fixed with `(_|$)` + longest-match selection.

Both are ad-hoc patches. A unified fix should be applied on `develop` before merging either branch.

## Proposed fix

Replace the substring grep with a proper word-boundary regex and longest-match resolution:

```r
parameter <- model_pars[unlist(lapply(
  paste0("(^|_)", model_pars, "(_|$)"),
  grepl, x = spar
))]
if (length(parameter) > 1) {
  parameter <- parameter[which.max(nchar(parameter))]
}
```

[...]

PR #383, opened 2026-07-07
Fix substring collision in summary row selection (#379)

### LINKED ISSUE #379: `print.bmmsummary` selects coefficient rows by a colliding substring match
`print.bmmsummary` picks which coefficient rows to show with `grepl(paste0(par, "_"), rownames(x$fixed))` ([summary.R:90-92](R/summary.R#L90-L92)). The match is an unanchored substring, so `"a_"` matches `"kappa_Intercept"`, and a model with parameters `a` and `kappa` (e.g. imm) can keep or drop the wrong rows. This is the same substring-collision class as #354, sitting in the summary path.

PR #369 fixes the single-row `apply()` crash in this exact block but keeps the `grepl`, so this half stays open. The fix is to tokenize the brms coefficient prefix and compare exactly — `sub("_.*$", "", rownames(x$fixed))` against the parameter set — rather than matching a substring. For every currently shipping model the printed row set is unchanged (verify against sdm/imm/mixture3p/m3). Best folded into #369 while that code is already being touched.

GidonFrischkorn:
`print.bmmsummary` picked coefficient rows with an unanchored `grepl("a_", ...)`, so `"a_"` also matched `"kappa_Intercept"` — a model with both `a` and `kappa` (e.g. imm) could keep the wrong rows. It now tokenizes the brms prefix and compares it exactly against the parameter set.

Dropping the `sapply`/`apply` step also fixes the single-row crash from #369, so both are resolved here. Printed output is unchanged for every shipping model (verified against the sdm and m3 fixtures). Added a regression test that fails under the old substring match.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ### Bug fixes
 * Fix `swald_lccdf()` returning incorrect log-survival probability when response time equals non-decision time in the **cswald** model. Previously returned `-Inf` instead of `0` (log of survival = 1) (#348).
+* Fix `print()` for model summaries selecting regression-coefficient rows by an unanchored substring match, so a parameter such as `a` could pull in rows of another parameter like `kappa` (e.g. in the **imm** model). Rows are now matched on the exact parameter prefix. This also fixes a crash when only a single coefficient row is shown (#379, #369).
 
 ### Other changes
 * The **ddm** model supports both `cmdstanr` and `rstan` backends. Previously, `cmdstanr` was required.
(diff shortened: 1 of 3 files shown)


GidonFrischkorn:
To be clear about scope: this does not currently mis-print any shipping model. The only rows whose prefix is not a printed parameter are the internal `kappa2`/`mu2` helpers in the circular mixture models, and no parameter is a suffix of those (they end in a digit), so the collision never changes the output.

So the goal is rather to harden the summary row selection against parameter-name substring collisions (#379, #369)

`print.bmmsummary` selected coefficient rows with an unanchored `grepl(paste0(par, "_"), ...)`, so `"a_"` also matches inside `"kappa_Intercept"`. This is the same substring-collision class as #354.

The change tokenizes the brms prefix and compares it exactly against the parameter set, which is simpler and removes the fragility before a future model trips it. Dropping the `sapply`/`apply` step also fixes the single-row crash from #369, which is genuinely reachable, so both are resolved here.

Added a regression test on the selection logic that fails under the old substring match.

